PR #1, opened 2023-04-18
sinol-make file structure and run command

MasloMaslane:
What's left to do:
- [x]  compile programs without makefile
~~check if dependencies are installed for run command (time and timeout)~~
~~add measuring time and memory with python as an option (do we want this?)~~


twalen (approved):
Just approving, if you would like to have normal review, please let me know.

tonowak:
> Just approving, if you would like to have normal review, please let me know.

While I plan to review the project in the context of added functionality, decisions and etc, I don't have much experience with Python. I would appreciate if you could make reviews :smile:

tonowak:
> add measuring time and memory with python as an option (do we want this?)

That depends on the existing issues with `oiejq`/`sio2jail`. One of the next steps would be to write tests (e.g. in GitHub Actions) to check whether the tool works correctly on the popular operating systems. If we won't find any issues, there is no need to have a separate measuring option (especially that I want to force the usage of `oiejq`). If we'll have issues, then we can consider implementing the `time`.

README.md
```diff
@@ -0,0 +1,18 @@
+# sinol-make
```
tonowak, line 1: Currently the project is not ready to be used, thus it might be best to inform about this in the `README.md` and update it when we'll have clear installation steps.

pyproject.toml
```diff
@@ -0,0 +1,31 @@
...
+	"requests",
+  "PyYAML",
+]
+
+[project.urls]
+"Homepage" = "https://github.com/sio2project/sinol-make"
+"Bug Tracker" = "https://github.com/sio2project/sinol-make/issues"
+
+[project.scripts]
+sinol-make = "sinol_make:main"
```
tonowak, line 31: Missing newline at the end of the file. You probably can try to configure your text editor to add them automatically.

MasloMaslane:
> > What's left to do:
> 
> Let's try to implement then in separate PRs.

You mean that compiling without makefiles should be in seperate PR? There is some unfinished code for compiling [here](https://github.com/sio2project/sinol-make/pull/1/files#diff-311b93157ef6618fbf24a334df308543b7d33423611e360b6edea8e809f6f6fc)

pyproject.toml
```diff
@@ -0,0 +1,31 @@
+[build-system]
+requires = ["setuptools>=67.0"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "sinol-make"
+version = "0.0.1"
+authors = [
+  { name="Mateusz Masiarz", email="m.masiarz@fri.edu.pl" },
+  { name="Tomasz Nowak", email="tomek.nowak.gm.16@gmail.com" }
```
tonowak, line 10: I do not want to take credits for your code. This project will be your property and it is more beneficial for you to be the only author (to write about this in the CV, etc). If there is a separate field for maintainers or something like that, it's better to put me there.

pyproject.toml
```diff
@@ -0,0 +1,31 @@
...
+requires-python = ">=3.7"
+classifiers = [
+    "Programming Language :: Python :: 3",
+    "License :: OSI Approved :: MIT License",
+    "Operating System :: OS Independent",
+]
+dependencies = [
+	"argparse",
+	"requests",
+  "PyYAML",
```
tonowak, line 23: Different indentations. You probably can configure your editor to be consistent.

pyproject.toml
```diff
@@ -0,0 +1,31 @@
...
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "sinol-make"
+version = "0.0.1"
+authors = [
+  { name="Mateusz Masiarz", email="m.masiarz@fri.edu.pl" },
+  { name="Tomasz Nowak", email="tomek.nowak.gm.16@gmail.com" }
+]
+description = "Tool for creating sio2 tasks"
```
tonowak, line 12:
```suggestion
description = "CLI tool for creating sio2 task packages"
```

pyproject.toml
tonowak, line 31: What this field means (I have no experience with `pip`)? Does it define which function to execute when used as a CLI tool?

src/sinol_make/commands/run/__init__.py
tonowak, line 1: Can you do a summary of the changes that were made here, compared to the original Bartosz's code? I'm not sure whether some functionality was added or only refactors were made.

src/sinol_make/config/__init__.py
```diff
@@ -0,0 +1,39 @@
+import yaml, os, subprocess
+
+class Config:
+	def __init__(self):
+		config_path = os.path.join(os.path.dirname(__file__), 'config.yml')
+		if not os.path.exists(config_path):
+			open(config_path, 'w').write(yaml.dump({}))
+			self.setup()
```
tonowak, line 8: I would say that the default behavior should be to panic -- our (current) assumption should be that the user is working on an existing repository with the task. It will change after adding a command for creating a directory with the task, but then still the tool should assume that the config exists.

src/sinol_make/config/__init__.py
```diff
@@ -0,0 +1,39 @@
...
+
+	def setup(self):
+		print("Configure your compilers:")
+		
+		self.set('c_compiler', self.__ask_value("C compiler", "gcc"))
+		self.set('cpp_compiler', self.__ask_value("C++ compiler", "g++"))
+		self.set('python_compiler', self.__ask_value("Python compiler", "python3"))
+		self.set('java_compiler', self.__ask_value("Java compiler", "javac"))
```
tonowak, line 16: Are there possible scenarios where the user would want something else?

src/sinol_make/util.py
```diff
@@ -0,0 +1,109 @@
...
+	def strip(tar):
+		l = len('oiejq/')
+		for member in tar.getmembers():
+			member.name = member.name[l:]
+			yield member
+
+	tar = tarfile.open('/tmp/oiejq.tar.gz')
+	tar.extractall(path=os.path.expanduser('~/.local/bin'), members=strip(tar))
+	tar.close()
+	os.remove('/tmp/oiejq.tar.gz')
```
tonowak, line 81: It is more convenient to have a command named `oiejq` instead of `oiejq.sh`, so we can rename the file to achieve this.

src/sinol_make/util.py
```diff
@@ -0,0 +1,109 @@
...
+			return True
+		else:
+			return False
+	
+	if check('oiejq.sh'):
+		return 'oiejq.sh'
+	elif check(os.path.expanduser('~/.local/bin/oiejq.sh')):
+		return os.path.expanduser('~/.local/bin/oiejq.sh')
+	else:
+		return None
```
tonowak, line 103: That is some code behavior that the user doesn't know about and can possibly frustrate the user (e.g. when one would put `oiejq` and modify it manually to achieve something, the user might not know that then `sinol-make` will use it). It's better to make it clear via CLI arguments. We can e.g. by default assume that `oiejq` is in the local bin directory and we possibly can overwrite this behavior by providing an argument to change the path (but in this case, we should have a clear command description that informs the user that by default the tool downloads and puts `oiejq` to local bin).

tonowak (commented):
Overall looks good!

tonowak:
> I think that oiejq only works on linux. I remember trying to compile it on macOS, but I didn't succeed. Maybe after changing some code it may work?

From what I remember, Marek Sokołowski might be a good person to ask about MacOS stuff.

pyproject.toml
MasloMaslane: It means that when used as CLI it will run function `main` from `__init__.py` in root of sinol-make

src/sinol_make/config/__init__.py
MasloMaslane: This config is for global settings, for example the names of compiler programs etc. Maybe we should rename the file to GlobalConfig or something like that?

src/sinol_make/config/__init__.py
MasloMaslane: Yes, for example on MacOS `g++` is clang compiler. The normal GNU C++ compiler is named g++-12 for version 12 or g++-9 for version 9.

pyproject.toml
tonowak: I see, thanks!

src/sinol_make/config/__init__.py
MasloMaslane: Yes, I wanted to store there `c_compiler`, `cpp_compiler`, ... and maybe other things in the future

src/sinol_make/config/__init__.py
tonowak: Maybe then we can try to deduce the compiler the tool should use, instead of asking the user to provide explicitly this information? Though then it might be beneficial to overwrite something, but I would argue that it should be a rare case and an argument/flag might be better suited.

src/sinol_make/config/__init__.py
tonowak: Having multiple config files might be harmful -- one might forget where one was or what is its content, the configs maybe would have to be copy-pasted between task packages, it pollutes the file structure with not relevant files for the task, etc. It also slows down the first usage of the tool -- one has to write a bunch of config values before running it for the first time. Thus we have to be careful with creating config files. Is it really needed? If someone would want to temporarily overwrite a config value, maybe a commandline argument with sensible default values might be better? Will we really have the need to put in this config other kind of values?

pyproject.toml
MasloMaslane: 3119b831ce78324e3399fd78ccc91b64245bb9a7

pyproject.toml
MasloMaslane: 3119b831ce78324e3399fd78ccc91b64245bb9a7

pyproject.toml
MasloMaslane: 3119b831ce78324e3399fd78ccc91b64245bb9a7

pyproject.toml
MasloMaslane: 3119b831ce78324e3399fd78ccc91b64245bb9a7

src/sinol_make/commands/run/__init__.py
MasloMaslane: 3119b831ce78324e3399fd78ccc91b64245bb9a7

tonowak: Thanks! Now I know what to review and look for :)

As we're using `git`, we don't have to write the file's history in the file. I don't have a problem with it if you're going to keep the changelog up-to-date with your changes, but often people fail doing it, so I ask you to reconsider it.

pyproject.toml
```diff
@@ -0,0 +1,33 @@
...
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "sinol-make"
+version = "0.0.1"
+authors = [
+  { name="Mateusz Masiarz", email="m.masiarz@fri.edu.pl" }
+]
+maintainers = [
+  { name="Tomasz Nowak", email="tomek.nowak.gm.16@gmail.com" }
```
tonowak, line 12:
```suggestion
  { name="Tomasz Nowak", email="tomasz.nowak@tonowak.com" }
```

src/sinol_make/commands/run/__init__.py
```diff
@@ -0,0 +1,509 @@
...
+							% (output_file, self.get_output_file(test))):
+					result["Status"] = "WA"
+				elif result["Time"] > time_limit:
+					result["Status"] = "TL"
+				elif result["Memory"] > memory_limit:
+					result["Status"] = "ML"
+			else:
+				result["Status"] = result["Status"][:2]
+			return result
+		elif self.args.time_tool == 'time':
```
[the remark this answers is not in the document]
MasloMaslane: OK, so I'll remove this for now

MasloMaslane:
Also, which license do we want to use?

tonowak:
The same as `sio2`, I guess.

MasloMaslane:
9d1b07a4a7d8424cfc1372cb672141149726017b I removed global config and finished compilation without makefiles

tonowak:
Nice! I'll try to review it today.

src/sinol_make/commands/run/__init__.py
```diff
@@ -0,0 +1,513 @@
...
+		parser.add_argument('--oiejq_path', type=str,
+		      				help='path to oiejq executable (default: `~/.local/bin/oiejq`)')
+		parser.add_argument('--c_compiler', type=str,
+		    				help='C compiler to use (default for Linux and Windows: gcc, default for Mac: gcc-{9-12})')
+		parser.add_argument('--cpp_compiler', type=str,
+		    				help='C++ compiler to use (default for Linux and Windows: g++, default for Mac: gcc-{9-12})')
+		parser.add_argument('--python_interpreter', type=str,
+		    				help='Python interpreter to use (default: python3)')
+		parser.add_argument('--java_compiler', type=str,
+		    				help='Java compiler to use (default: javac)')
```
tonowak, line 59: Maybe there is a way to set a default value to those flags, that is, when the user doesn't provide any value, it assumes the default one.

src/sinol_make/helpers/compiler.py
tonowak, line 1: There is a lot of code duplication. Maybe we can make this file shorter and more readable?

src/sinol_make/helpers/compiler.py
```diff
@@ -0,0 +1,109 @@
...
+		if not check_if_installed('gcc'):
+			print(util.bold(util.color_red('C compiler \'gcc\' not found. You can set the C compiler with --c_compiler flag')))
+			exit(1)
+		else:
+			return 'gcc'
+	elif sys.platform == 'darwin':
+		for i in range(9, 13):
+			compiler = 'gcc-' + str(i)
+			if check_if_installed(compiler):
+				return compiler
```
tonowak, line 40: What if it's neither `win32`, `linux` and `darwin`?

MasloMaslane: here are possible values of sys.platform https://docs.python.org/3.9/library/sys.html#sys.platform

MasloMaslane: I forgot about cygwin, i don't think that aix is used anywhere

src/sinol_make/commands/run/__init__.py
MasloMaslane: There is, but what should be the default value for c++ compiler, since it can be g++ on linux or g++-{9-12} on MacOS?

src/sinol_make/helpers/compiler.py
tonowak: In this case, we can add an `else assert` with a message that other operating systems are not used widely or not supported/documented, with the link you sent here.

src/sinol_make/commands/run/__init__.py
tonowak: I don't know how the Python CLI library works, but in some languages/libraries it's possible to send a function instead of a value, which evaluates and returns a default value.

src/sinol_make/helpers/compiler.py
MasloMaslane: OK, there are a bit more possible values https://stackoverflow.com/questions/446209/possible-values-from-sys-platform

src/sinol_make/commands/run/__init__.py
MasloMaslane: And if there is no c++ compiler found should we just return g++?

tonowak: As we assume that this case shouldn't happen, an error with a suitable message might be better (explaining what the tool tried to do and that it's possible to manually set the command).

src/sinol_make/config/__init__.py
MasloMaslane: Removed global config

src/sinol_make/util.py
MasloMaslane: So by default when oijeq is not found in `~/.local/bin` we should download it?

src/sinol_make/commands/run/__init__.py
MasloMaslane: f2b03c5f50acc72591fdf840326c3a669f17eec1

src/sinol_make/helpers/compiler.py
MasloMaslane: Is it better now?

README.md
MasloMaslane: ebc5c617e28d00469f54f68baab4408506c8c5fe

src/sinol_make/helpers/compile.py
tonowak, line 1: The functions here have similar content. Maybe it would be shorter/more readable to remove code duplication? (I haven't put much thought into this -- it's just a suggestion to consider).

src/sinol_make/commands/run/__init__.py
```diff
@@ -0,0 +1,542 @@
...
+							for program in programs]
+		compiled_commands = zip(programs, program_executables, compilation_results)
+		names = programs
+		self.perform_executions(compiled_commands, names, programs, self.args.program_report)
+	
+	
+	def run(self, args):
+		if not util.check_if_project():
+			print(util.bold(util.color_red('You are not in a project directory.')))
+			exit(1)
```
tonowak, line 434: I like that! Let's also add an information (doesn't have to also be in bold/red) why the tool thinks that we are not in a project directory -- e.g. that it didn't find `config.yml` in the current working directory.

src/sinol_make/util.py
```diff
@@ -0,0 +1,111 @@
...
+		return False
+	if check_oiejq():
+		return True
+
+	if not os.path.exists(os.path.expanduser('~/.local/bin')):
+		os.makedirs(os.path.expanduser('~/.local/bin'))
+
+	request = requests.get('https://oij.edu.pl/zawodnik/srodowisko/oiejq.tar.gz')
+	if request.status_code != 200:
+		raise Exception('Couldn\'t download oiejq (oij.edu.pl status code: ' + str(request.status_code) + ')')
```
tonowak, line 72: Let's also add the link from which we tried to download, so the user can check whether it's something wrong with the site or with the tool.

src/sinol_make/commands/run/__init__.py
```diff
@@ -0,0 +1,542 @@
...
+		parser.add_argument('--validate_subtasks',
+							dest='validate_subtasks', action='store_true',
+							help='validate subtasks defined in config.yml')
+		parser.add_argument('--validate_programs',
+							dest='validate_programs', action='store_true',
+							help='validate programs in prog/ directory')
+		parser.add_argument('--subtask_report', type=str,
+							help='file to store report from subtask validation (in markdown)')
+		parser.add_argument('--program_report', type=str,
+							help='file to store report from program executions (in markdown)')
```
tonowak, line 49: What do you think about removing those arguments? (open for discussion)

My though process:
- `--verbose` is not clear what it means (it usually means that the CLI tool prints more data about what it's currently doing) and I doubt that the current name is intuitive with the current behavior (it expands the printed table). Because sometimes a solution isn't working as expected, the user behavior would probably be "running it without this flag, then seeing that something is wrong and rerunning the tool with this flag", which is a waste of time for the user when the solutions don't check quickly. Instead, I think that it's reasonable to print both a short table (like without this flag) and under it print a long table (like with this flag), so that the user can choose whether he wants to look into the short table or it's necessary to look into the results of the long table. Printing both tables at once can be done in a separate PR, it's not important to do it now.
- `--show_memory` should be enabled by default and there isn't really any reason why someone would want to disable it, besides when `oiejq` doesn't work with printing the memory. But in such a case, it might be better to detect that printing the memory doesn't work and somehow cope with that.
- `--validate_subtasks` and `--validate_programs` should always be enabled, the tool should first run all the solutions and then check whether a solution scored a different number of points than expected.
- `--subtask_report` and `--program_report` probably should be merged together and enabled by default, so that the user has the latest report in a saved file (just like `raport.html` in sinol packages -- we can try to replace this file with those reports). That would also solve the fact of missing/forgetting creating the reports (though, my proposal of `sinol_expected_scores` already mostly solves this issue, thus I'm leaving it open for discussion).

I'm basing my propositions on two rules:

It's good to polish the available commands and their names now, when we haven't yet written tests for them (so that we won't have to change them later). If we're going to need one of the flags, we can always add it later -- from the user experience, it's much easier to add functionality instead of changing it. Polishing the descriptions of command can be made later (which we will do -- clear command descriptions and a clear `--help` message is critical for pleasant usage of the tool).

The user should have reasonable default arguments that checks whether the user makes no mistakes as much as possible. If there is no reason to provide the user with a mechanism that disables one of the checks, we shouldn't do it. Thanks to that, the user will have to look into more detail into why a check failed, instead of ignoring it and running the tool without a check.

tonowak (commented):
Just some small comments. I think we're going to merge this PR as soon as they are resolved :rocket:

src/sinol_make/commands/run/__init__.py
MasloMaslane: ``verbose``: I agree that verbose here doesn't mean what it usually means, we should delete id

MasloMaslane: ``show_memory``: I agree, but is there a possibility that oiejq doesn't measure memory?

tonowak: > show_memory: I agree, but is there a possibility that oiejq doesn't measure memory?

Yes, on Arch Linux often there are problems with it. We'll see whether that's indeed the case in step 1c (tests).

tonowak: > but I don't think we should replace raport.html

What ideas do you have then?

src/sinol_make/commands/run/__init__.py
```diff
@@ -0,0 +1,542 @@
...
+		parser.add_argument('--oiejq_path', type=str,
+		      				help='path to oiejq executable (default: `~/.local/bin/oiejq`)')
+		parser.add_argument('--c_compiler', type=str, default=compiler.get_c_compiler(),
+		    				help='C compiler to use (default for Linux and Windows: gcc, default for Mac: gcc-{9-12})')
+		parser.add_argument('--cpp_compiler', type=str, default=compiler.get_cpp_compiler(),
+		    				help='C++ compiler to use (default for Linux and Windows: g++, default for Mac: gcc-{9-12})')
+		parser.add_argument('--python_interpreter', type=str, default=compiler.get_python_interpreter(),
+		    				help='Python interpreter to use (default: python3)')
+		parser.add_argument('--java_compiler', type=str, default=compiler.get_java_compiler(),
+		    				help='Java compiler to use (default: javac)')
```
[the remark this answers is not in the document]
MasloMaslane: c17ac3213a8dfbe54ec7c935eec0593988b584b7

src/sinol_make/commands/run/__init__.py
MasloMaslane: Maybe we could change the current behavior so that it generates ``raport.html``?

tonowak: Let me know what you think about this approach: when the user runs the tool and it prints a table (that shows for each solution got how many points on each test), assuming that the user checks all of the solutions on all of the tests, the tool can automatically save the same table to `raport.html` (if we're going to make so that the tool shows both short and long tables during `sinol-make run`, we can also put both the tables into this file).

MasloMaslane: Yes, I think that's a good idea

MasloMaslane: By default when ``--validate_subtasks`` and ``--validate_programs`` are both not set, the program runs subtasks and programs. Should I leave it like that, or do we want to delete these flags?

tonowak: My plan was to remove the "subtasks" check and instead have the one I proposed in the docs. Then the flags won't be at all needed. I'm open for suggestions though.

MasloMaslane: New subtask config: ac0accae2faedc5c318dcea52fec683f9721aba0
Suggest subtask configuration: 63ac565f28695a63e4c5679382a33d08d91064a6

MasloMaslane:
Config suggestions was moved to #3, so I think we can merge this. @tonowak what do you think?

tonowak (approved):
Looks good, I'm giving approve so that you can merge (if you think that the TODO comments are not enough and you'll forget about changing something, you can create an issue). Though, before merging, please resolve the last open comment.

src/sinol_make/helpers/compile.py
```diff
@@ -0,0 +1,44 @@
...
+def compile(program, output, compilers = None, compile_log = None):
+	"""
+	Compile a program
+	compilers - A dictionary of compilers to use. If not set, the default compilers will be used
+	"""
+	ext = os.path.splitext(program)[1]
+	arguments = []
+	if ext == '.cpp':
+		arguments = [compilers['cpp_compiler_path'] or compiler.get_cpp_compiler_path(), program, '-o', output] + \
+						'-O3 -lm -Werror -Wall -Wextra -Wshadow -Wconversion -Wno-unused-result -Wfloat-equal'.split(' ')
```
Kostero, line 14: can we make this as a variable/argument and then just pass `C_FLAGS`?

MasloMaslane: Maybe we should add flags to config.yml and if none are set then the code should use less strict flags then those that are now used?

MasloMaslane: Or by default the flags should be the same as those that oioioi uses

src/sinol_make/commands/run/__init__.py
```diff
@@ -0,0 +1,603 @@
...
+					exit(1)
+				else:
+					print(util.bold(util.color_red('Couldn\'t find a C++ compiler. Tried g++. Try specifying a compiler with --cpp_compiler_path.')))
+					exit(1)
+			elif ext == '.py' and args.python_interpreter_path is None:
+				print(util.bold(util.color_red('Couldn\'t find a Python interpreter. Tried python3. Try specifying an interpreter with --python_interpreter_path.')))
+				exit(1)
+			elif ext == '.java' and args.java_compiler_path is None:
+				print(util.bold(util.color_red('Couldn\'t find a Java compiler. Tried javac. Try specifying a compiler with --java_compiler_path.')))
+				exit(1)
```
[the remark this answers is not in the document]
MasloMaslane: Is it better now?

src/sinol_make/helpers/compile.py
tonowak: It would indeed be beneficial to be able to pass custom flags, e.g. `-fsanitize=address,undefined` or `-D_GLIBCXX_DEBUG`, or temporarily disable one of the warning flags.

But then, the calculated times won't necessarily be the same as the ones on sio2 and it might also trigger the `sinol_expected_scores` mechanism. If we want to have the ability to specify flags, it should give a warning that the times might not be the same and ignore the `sinol_expected_scores`. Probably later we would want to have a `sinol-make verify` that runs each command and checks whether they work as expected (e.g. the input files haven't changed their hashes, they satisfy the `inwer`, the output files are correct, the `sinol_expected_scores` is set as expected, etc) and in this command we wouldn't want the user to specify the custom flags.

To do it perfectly we might need to put into it more thought and the implementation won't be trivial. I think, at this point, it's better to create an Issue for later, focus on merging this PR and adding non-necessary functionality after writing basic tests.

MasloMaslane: I created the issue #4, so I think that we can merge this PR

PR #2, opened 2023-04-19
Add possibility to measure performance with `time`

MasloMaslane:
This PR adds option to measure time and memory usage with `time`. We have to decide whether we want to add this.
What's left to do:
- [x] check if `time` and `timeout` is installed


tonowak:
We can wait with this until we'll have tests for different operating systems.

README.md
```diff
@@ -5,3 +5,16 @@ Currently in development.
...
+## MacOS
+### Measuring time with oiejq
+Not available
+### Measuring time with time command
+Install those packages with brew: \
+```brew install coreutils gnu-time```
+
+## Linux
+### Measuring time with oiejq
+Download oiejq from [here](https://oij.edu.pl/zawodnik/srodowisko/oiejq.tar.gz) and add oiejq.sh to `$PATH`
```
tonowak, line 17: Those steps are not true anymore, right?

src/sinol_make/commands/run/__init__.py
```diff
@@ -201,37 +203,72 @@ def execute(self, execution):
...
+		elif self.args.time_tool == 'time':
+			if sys.platform == 'darwin':
+				command = 'launchctl limit memlock %s; gtimeout -k %ds %ds gtime -f "%%U\\n%%M\\n%%x" -o %s %s <%s >%s' \
+					% (math.ceil(memory_limit) * 1024, hard_time_limit_in_s,
+						hard_time_limit_in_s, result_file, program, test, output_file)
+			elif sys.platform == 'linux':
+				command = 'ulimit -v %s; timeout -k %ds %ds time -f "%%U\\n%%M\\n%%x" -o %s %s <%s >%s' \
+					% (math.ceil(memory_limit), hard_time_limit_in_s,
+						hard_time_limit_in_s, result_file, program, test, output_file)
+
```
tonowak, line 247: What about `windows` etc? While it might not be necessary (because such a case won't happen), it makes the code more clear when there is an `else` with an assert/throw with an error message that describes why this case won't happen.

src/sinol_make/commands/run/__init__.py
```diff
@@ -201,37 +203,72 @@ def execute(self, execution):
...
+				if os.system("diff -q -Z %s %s >/dev/null"
+							% (output_file, self.get_output_file(test))):
+					result["Status"] = "WA"
+				elif result["Time"] > time_limit:
+					result["Status"] = "TL"
+				elif result["Memory"] > memory_limit:
+					result["Status"] = "ML"
+			return result
+		elif self.args.time_tool == 'python':
+			pass
```
tonowak, line 271: The code becomes quite lengthy and they are implementation details, so it might be better to split the existing code into smaller functions.

README.md
MasloMaslane: I don’t think that in #1 I added automatic oiejq download

tonowak: Oh, I see, sorry. Then I hope someday you'll create a PR with it :)

src/sinol_make/commands/run/__init__.py
MasloMaslane: It should be better now 80b92ca6682a772342abb132bea23d0a09d3cb92

src/sinol_make/commands/run/__init__.py
MasloMaslane: Now the code throws an exception d55450a7229741e6e1f0a05e64f91a9bd1b9bfb4

MasloMaslane:
I also added code to check whether time and timeout is installed. 867f3132b80bec8fbadcbf0b3ae184b368a76fb9

tonowak:
This PR also needs a rebase unfortunately.

PR #3, opened 2023-04-28
Suggest sinol_expected_scores if none found in config.yml

MasloMaslane:
### What has been done:
- rename subtasks to sinol_expected_scores
- sinol_expected_scores describes what results should given program have
- if sinol_expected_scores is not found in config, all programs are run on all tests and sinol-make suggests new config with sinol_expected_scores

### To do:
- [x] refactor code

### DIFF
--- a/src/sinol_make/commands/run/__init__.py
+++ b/src/sinol_make/commands/run/__init__.py
@@ -499,13 +499,13 @@ class Command(BaseCommand):
 
 		def warn_if_not_empty(set, message):
 			if len(set) > 0:
-				print(util.warning(message), end='')
+				print(util.warning(message + ": "), end='')
 				print(util.warning(", ".join([str(x) for x in set])))
 
-		warn_if_not_empty(added_solutions, "Solutions were added: ")
-		warn_if_not_empty(removed_solutions, "Solutions were removed: ")
-		warn_if_not_empty(added_groups, "Groups were added: ")
-		warn_if_not_empty(removed_groups, "Groups were removed: ")
+		warn_if_not_empty(added_solutions, "Solutions were added")
+		warn_if_not_empty(removed_solutions, "Solutions were removed")
+		warn_if_not_empty(added_groups, "Groups were added")
+		warn_if_not_empty(removed_groups, "Groups were removed")
 
 		for change in changes:
 			print(util.warning("Solution %s passed group %d with status %s while it should pass with status %s." %
--- a/src/sinol_make/commands/run/__init__.py
+++ b/src/sinol_make/commands/run/__init__.py
@@ -6,7 +6,7 @@ from sinol_make.interfaces.BaseCommand import BaseCommand
 from sinol_make.interfaces.Errors import CompilationError
 from sinol_make.helpers import compile, compiler
 import sinol_make.util as util
-import yaml, os, collections, sys, re, math
+import yaml, os, collections, sys, re, math, dictdiffer
 import multiprocessing as mp
 
 class Command(BaseCommand):
@@ -430,6 +430,14 @@ class Command(BaseCommand):
 		return self.perform_executions(compiled_commands, names, programs, self.args.program_report)
 
 
+	def print_expected_scores(self, expected_scores):
+		print("sinol_expected_scores:")
+		for program in expected_scores.keys():
+			print("  %s:" % program)
+			print("    points: %d" % expected_scores[program]["points"])
+			print("    expected: " + str(expected_scores[program]["expected"]))
+
+
 	def validate_expected_scores(self, results, programs):
 		if "sinol_expected_scores" not in self.config.keys():
 			print(util.bold("Suggested expected scores description:"))
@@ -466,87 +474,107 @@ class Command(BaseCommand):
 					"points": self.calculate_points(results[program])
 				}
 
-			expected_scores = {} # Expected scores from config only for programs and tests specified
-			error = False
-			new_programs = []
-			deleted_programs = []
-			for program in programs:
-				# Check if there is a new program
-				if program not in self.config["sinol_expected_scores"]:
-					if program in new_programs:
-						continue
+			expected_scores = {} # Expected scores from config with only programs and groups that were run
+			for program in results.keys():
+				if program in self.config["sinol_expected_scores"]:
+					expected_scores[program] = {
+						"expected": {},
+						"points": 0
+					}
 
-					print(util.warning(f'There seems to be a new program {program}'))
-					new_programs.append(program)
-					error = True
-					continue
-
-				# Get expected scores for given tests from config
-				expected_scores[program] = {
-					"expected": {},
-					"points": 0
-				}
-				for test in self.tests:
-					group = self.get_group(test)
-					if group not in expected_scores[program]["expected"]:
-						expected_scores[program]["expected"][group] = self.config["sinol_expected_scores"][program]["expected"][group]
-
-				expected_scores[program]["points"] = self.calculate_points(expected_scores[program]["expected"])
-
-			# Check if any program was deleted
-			all_programs = self.get_programs(None)
-			for program in self.config["sinol_expected_scores"].keys():
-				if program not in all_programs:
-					print(util.warning(f'Program {program} was deleted.'))
-					error = True
-					deleted_programs.append(program)
-
-			if new_expected_scores != expected_scores:
-				# Programs both in config and new results
-				common_programs = [program for program in programs if program not in new_programs]
-
-				for program in common_programs:
-					for group, result in new_expected_scores[program]["expected"].items():
-						# Check if there is a new group
-						if group not in expected_scores[program]["expected"]:
-							print(util.warning(f'There seems to be a new group {group} for program {program}.'))
-							error = True
-							continue
-
-						# Check if programs passed with expected output
-						if result != expected_scores[program]["expected"][group]:
-							print(util.warning(f'Program {program} passed group {group} with status {result}, '
-								f'while it should pass with status {expected_scores[program]["expected"][group]}.'))
-							error = True
-
-			if self.args.apply_suggestions and error:
-				# Generate new config
-				expected_scores = self.config["sinol_expected_scores"]
-				for program in new_expected_scores.keys():
-					if program not in expected_scores: # if there is a new program found
-						expected_scores[program] = {
-							"expected": {},
-							"points": 0
-						}
-					for group, result in new_expected_scores[program]["expected"].items(): # Update expected results
-						expected_scores[program]["expected"][group] = result
+					for group in results[program].keys():
+						if group in self.config["sinol_expected_scores"][program]["expected"]:
+							expected_scores[program]["expected"][group] = self.config["sinol_expected_scores"][program]["expected"][group]
 
 					expected_scores[program]["points"] = self.calculate_points(expected_scores[program]["expected"])
 
-				# Remove deleted programs
-				for program in deleted_programs:
-					del expected_scores[program]
+			print(util.bold("Expected scores from config:"))
+			self.print_expected_scores(expected_scores)
+			print(util.bold("\nExpected scores based on results:"))
+			self.print_expected_scores(new_expected_scores)
 
-				self.config["sinol_expected_scores"] = expected_scores
-				with open(os.path.join(os.getcwd(), "config.yml"), "w") as f:
-					yaml.dump(self.config, f, default_flow_style=False)
-				print(util.info("Suggested expected scores description was applied."))
-			elif error:
-				print(util.error("Expected scores description is not valid. Use --apply_suggestions to apply results."))
-				exit(1)
+			expected_scores_diff = dictdiffer.diff(expected_scores, new_expected_scores)
+			added_programs = set()
+			removed_programs = set()
+			added_groups = set()
+			removed_groups = set()
+
+			for type, field, change in list(expected_scores_diff):
+				if type == "add":
+					if field == '': # Programs were added
+						for program in change:
+							added_programs.add(program[0])
+					elif field[1] == "expected": # Groups were added
+						for group in change:
+							added_groups.add(group[0])
+				elif type == "change":
+					if field[1] == "expected": # Result for group changed
+						program = field[0]
+						group = field[2]
+						old_result = change[0]
+						group = change[1]
+
+						print(util.warning("Program %s passed group %d with status %s while it should pass with status %s." %
+			 								(program, group, group, old_result)))
+
+			# Only if sinol_make was run on all programs we should check for removed programs
+			if self.args.programs == None:
+				for program in self.config["sinol_expected_scores"].keys():
+					if program not in new_expected_scores.keys():
+						removed_programs.add(program)
+
+			# Only if sinol_make was run on all groups we should check for removed groups
+			if self.args.tests == None:
+				for program in self.config["sinol_expected_scores"].keys():
+					for group in self.config["sinol_expected_scores"][program]["expected"].keys():
+						if program in new_expected_scores and group not in new_expected_scores[program]["expected"].keys():
+							removed_groups.add(group)
+
+			if len(added_programs) > 0:
+				print(util.warning("New programs were added: "), end='')
+				print(util.warning(", ".join(added_programs)))
+			if len(removed_programs) > 0:
+				print(util.warning("Programs were removed: "), end='')
+				print(util.warning(", ".join(removed_programs)))
+
+			if len(added_groups) > 0:
+				print(util.warning("New groups were added: "), end='')
+				print(util.warning(", ".join([str(group) for group in added_groups])))
+			if len(removed_groups) > 0:
+				print(util.warning("Groups were removed: "), end='')
+				print(util.warning(", ".join([str(group) for group in removed_groups])))
+
+			if expected_scores == new_expected_scores:
+				print(util.info("Expected scores are correct!"))
 			else:
-				print(util.info("Expected scores are valid."))
+				if self.args.apply_suggestions:
+					expected_scores = self.config["sinol_expected_scores"]
 
+					for program in removed_programs:
+						del expected_scores[program]
+
+					for program in expected_scores:
+						for group in removed_groups:
+							if group in expected_scores[program]["expected"]:
+								del expected_scores[program]["expected"][group]
+						expected_scores[program]["points"] = self.calculate_points(expected_scores[program]["expected"])
+
+					for program in new_expected_scores.keys():
+						if program in expected_scores:
+							for group, group in new_expected_scores[program]["expected"].items():
+								expected_scores[program]["expected"][group] = group
+							expected_scores[program]["points"] = self.calculate_points(expected_scores[program]["expected"])
+						else:
+							expected_scores[program] = new_expected_scores[program]
+
+
+					self.config["sinol_expected_scores"] = expected_scores
+					with open(os.path.join(os.getcwd(), "config.yml"), "w") as f:
+						yaml.dump(self.config, f, default_flow_style=False)
+					print(util.info("Saved suggested expected scores description."))
+				else:
+					print(util.error("Use flag --apply_suggestions to apply suggestions."))
+					exit(1)
 
 	def run(self, args):
 		if not util.check_if_project():
@@ -671,6 +699,6 @@ class Command(BaseCommand):
 				print(util.warning('Use flag --apply_suggestions to apply suggestions.'))
 
 		programs = self.get_programs(self.args.programs)
-		self.validate_config(programs)
+		# self.validate_config(programs)
 		results = self.run_programs(programs)
 		self.validate_expected_scores(results, programs)
--- a/src/sinol_make/commands/run/__init__.py
+++ b/src/sinol_make/commands/run/__init__.py
@@ -186,7 +186,7 @@ class Command(BaseCommand):
 		output = os.path.join(self.EXECUTABLES_DIR, program)
 		try:
 			compile.compile(source_file, output, self.compilers, open(compile_log_file, "w"))
-			print(util.color_green("Compilation of file %s was successful."
+			print(util.info("Compilation of file %s was successful."
 							% self.extract_program_name(program)))
 			return True
 		except CompilationError as e:
@@ -423,11 +423,10 @@ class Command(BaseCommand):
 		names = programs
 		results = self.perform_executions(compiled_commands, names, programs, self.args.expected_scores_report)
 
-		if suggestions:
+		if "sinol_expected_scores" not in self.config.keys():
 			print(util.bold("Suggested expected scores description:"))
 			print("sinol_expected_scores:")
 
-			i = 1
 			new_config = self.config
 			new_config["sinol_expected_scores"] = {}
 
@@ -447,22 +446,49 @@ class Command(BaseCommand):
 				new_config["sinol_expected_scores"][program]["points"] = points
 
 			if self.args.apply_suggestions:
-				print("Do you want to apply these suggestions? (y/n) ", end="")
-				choice = input().lower()
-				if choice == "y" or choice == "yes" or choice == "":
-					with open(os.path.join(os.getcwd(), "config.yml"), "w") as f:
-						yaml.dump(new_config, f, default_flow_style=False)
-					print(util.color_green("Suggestions applied."))
-				else:
-					print("Suggestions not applied.")
+				with open(os.path.join(os.getcwd(), "config.yml"), "w") as f:
+					yaml.dump(new_config, f, default_flow_style=False)
+			else:
+				print(util.warning("Use flag --apply_suggestions to apply suggestions."))
+				exit(1)
 		else:
-			for program in self.config["sinol_expected_scores"]:
-				for group, expected_result in self.config["sinol_expected_scores"][program]["expected"].items():
-					if results[program][group] != expected_result:
-						print(util.error('Program %s will pass group %d with result %s (expected %s).' % (program, group, results[program][group], expected_result)))
-						exit(1)
+			new_config = self.config
+			error = False
 
-			print(util.color_green("Expected scores are valid."))
+			for program in results.keys():
+				if program not in self.config["sinol_expected_scores"]:
+					print(util.warning("Program %s is not defined in expected scores description." % program))
+					error = True
+					new_config["sinol_expected_scores"][program] = {
+						"expected": results[program],
+						"points": 0
+					}
+					continue
+
+				for group, result in results[program].items():
+					if group not in self.config["sinol_expected_scores"][program]["expected"]:
+						print(util.warning("Group %d is not defined for program %s in expected scores description." % (group, program)))
+						error = True
+						new_config["sinol_expected_scores"][program]["expected"][group] = result
+						new_config["sinol_expected_scores"][program]["points"] += self.scores[group] if result == "OK" else 0
+						continue
+					if self.config["sinol_expected_scores"][program]["expected"][group] != result:
+						print(util.warning("Program %s passed group %d with status %s, while it should pass with status %s."
+			 								% (program, group, result, self.config["sinol_expected_scores"][program]["expected"][group])))
+						error = True
+						new_config["sinol_expected_scores"][program]["expected"][group] = result
+						new_config["sinol_expected_scores"][program]["points"] += self.scores[group] if result == "OK" else 0
+						continue
+
+			if self.args.apply_suggestions and error:
+				with open(os.path.join(os.getcwd(), "config.yml"), "w") as f:
+					yaml.dump(new_config, f, default_flow_style=False)
+				print(util.info("Suggested expected scores description was applied."))
+			elif error:
+				print(util.error("Expected scores description is not valid. Use --apply_suggestions to apply results."))
+				exit(1)
+			else:
+				print(util.info("Expected scores are valid."))
 
 
 	def run_programs(self):
--- a/src/sinol_make/commands/run/__init__.py
+++ b/src/sinol_make/commands/run/__init__.py
@@ -133,8 +133,9 @@ class Command(BaseCommand):
 		else:
 			all_programs = []
 			for program in arg_problems:
-				if os.path.isfile(program):
-					all_programs.append(self.get_executable(program))
+				if not os.path.isfile(program):
+					raise Exception("Program %s does not exist" % program)
+				all_programs.append(self.get_executable(program))
 			return sorted(all_programs, key=self.get_program_key)
 
 
--- a/src/sinol_make/commands/run/__init__.py
+++ b/src/sinol_make/commands/run/__init__.py
@@ -487,15 +487,7 @@ class Command(BaseCommand):
 		)
 
 
-	def print_expected_scores_diff(self, validation_results: ValidationResult):
-		added_solutions = validation_results.added_solutions
-		removed_solutions = validation_results.removed_solutions
-		added_groups = validation_results.added_groups
-		removed_groups = validation_results.removed_groups
-		changes = validation_results.changes
-		expected_scores = validation_results.expected_scores
-		new_expected_scores = validation_results.new_expected_scores
-
+	def print_expected_scores_diff(self, diff: ValidationResult):
 		config_expected_scores = self.config["sinol_expected_scores"] if "sinol_expected_scores" in self.config else {}
 
 		def warn_if_not_empty(set, message):
@@ -503,16 +495,16 @@ class Command(BaseCommand):
 				print(util.warning(message + ": "), end='')
 				print(util.warning(", ".join([str(x) for x in set])))
 
-		warn_if_not_empty(added_solutions, "Solutions were added")
-		warn_if_not_empty(removed_solutions, "Solutions were removed")
-		warn_if_not_empty(added_groups, "Groups were added")
-		warn_if_not_empty(removed_groups, "Groups were removed")
+		warn_if_not_empty(diff.added_solutions, "Solutions were added")
+		warn_if_not_empty(diff.removed_solutions, "Solutions were removed")
+		warn_if_not_empty(diff.added_groups, "Groups were added")
+		warn_if_not_empty(diff.removed_groups, "Groups were removed")
 
-		for change in changes:
+		for change in diff.changes:
 			print(util.warning("Solution %s passed group %d with status %s while it should pass with status %s." %
 										(change.solution, change.group, change.result, change.old_result)))
 
-		if expected_scores == new_expected_scores:
+		if diff.expected_scores == diff.new_expected_scores:
 			print(util.info("Expected scores are correct!"))
 		else:
 			def delete_group(solution, group):
@@ -526,19 +518,19 @@ class Command(BaseCommand):
 
 
 			if self.args.apply_suggestions:
-				for solution in removed_solutions:
+				for solution in diff.removed_solutions:
 					del config_expected_scores[solution]
 
 				for solution in config_expected_scores:
-					for group in removed_groups:
+					for group in diff.removed_groups:
 						delete_group(solution, group)
 
-				for solution in new_expected_scores.keys():
+				for solution in diff.new_expected_scores.keys():
 					if solution in config_expected_scores:
-						for group, result in new_expected_scores[solution]["expected"].items():
+						for group, result in diff.new_expected_scores[solution]["expected"].items():
 							set_group_result(solution, group, result)
 					else:
-						config_expected_scores[solution] = new_expected_scores[solution]
+						config_expected_scores[solution] = diff.new_expected_scores[solution]
 
 
 				self.config["sinol_expected_scores"] = config_expected_scores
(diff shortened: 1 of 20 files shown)


MasloMaslane:
516a409fee060b92ad5240c63ac7a4562e5266a3 I did some refactoring.

src/sinol_make/commands/run/__init__.py
```diff
@@ -368,48 +377,98 @@ def print_view(output_file=None):
...
 
-	def validate_subtasks(self):
-		print("Validating subtasks...")
-		if 'subtasks' not in self.config.keys():
-			print(util.bold(util.color_red('Subtasks description not defined in config.yml.')))
-			exit(1)
+	def validate_expected_scores(self):
+		print("Validating expected scores...")
+		if 'sinol_expected_scores' not in self.config.keys():
+			print(util.bold(util.color_yellow('Expected scores description not defined in config.yml. ' \
```
tonowak, line 383: The "print bold color and exit" pattern shows *very often* in this code, but it's merely an implementation detail and is not worth my attention during review. I suggest to move it to a separate function (e.g. one that produces an error and one that produces a warning). That will shorten the code an allow me to more quickly read the code.

src/sinol_make/commands/run/__init__.py
```diff
@@ -368,48 +377,98 @@ def print_view(output_file=None):
...
-				exit(1)
+		if suggestions:
+			programs = self.get_solutions()
+		else:
+			for prog in self.config["sinol_expected_scores"]:
+				score_checksum = 0
+				for group, expected_result in self.config["sinol_expected_scores"][prog]["expected"].items():
+					if group not in self.scores.keys() or group == 0:
+						print(util.bold(util.color_red('Group %d was not defined.' % group)))
+						exit(1)
```
tonowak, line 400: Should there be an error when group 0 exists? Usually group 0 is the sample tests and usually it exists.

src/sinol_make/commands/run/__init__.py
```diff
@@ -368,48 +377,98 @@ def print_view(output_file=None):
...
+					if expected_result not in ["TLE", "MLE", "RTE", "WA", "OK"]:
+						print(util.bold(util.color_red('Expected result for group %d is not valid.' % group)))
+						exit(1)
+
+					if expected_result == "OK":
+						score_checksum += self.scores[group]
+
+				score_expected = self.config["sinol_expected_scores"][prog]["points"]
```
tonowak, line 408: Can you write me an example of a `sinol_expected_scores`, so that I can confirm that the code does what it should, instead of guessing what it does?

src/sinol_make/commands/run/__init__.py
```diff
@@ -368,48 +377,98 @@ def print_view(output_file=None):
...
+
+			if self.args.apply_suggestions:
+				print("Do you want to apply these suggestions? (y/n) ", end="")
+				choice = input().lower()
+				if choice == "y" or choice == "yes" or choice == "":
+					with open(os.path.join(os.getcwd(), "config.yml"), "w") as f:
+						yaml.dump(new_config, f, default_flow_style=False)
+					print(util.color_green("Suggestions applied."))
+				else:
+					print("Suggestions not applied.")
```
tonowak, line 458: While the "ask for confirmation" is a good idea (and the implementation seems to be nice), we first need to think about how the users will use the tool and what is one's expectations with each flags.

I think that the users that use `--apply-suggestion` would use it when they are pretty much sure that the suggestion is at it is, or they want to ignore the whole suggestion mechanism (e.g. on the initial stages of creating a task package).

On the contrary, when used without this flag should serve as a confirmation that e.g. the saved scores is wrong and in that case probably the user would want to see the difference (what was saved vs what should be there).

Having those use cases in mind, it would be reasonable to print "what was saved vs what should be there" if something is wrong (bonus points for first printing all of than and only then exiting) and the `--apply-suggestions` should automatically apply them and print what was changed & how it was changed.

Let me know what you think.

src/sinol_make/commands/run/__init__.py
```diff
@@ -368,48 +377,98 @@ def print_view(output_file=None):
...
+		else:
+			for prog in self.config["sinol_expected_scores"]:
+				program = os.path.join(self.EXECUTABLES_DIR, self.config["sinol_expected_scores"][prog]["validator"])
+				compiled_commands.append((prog, program, True))
+			names = list(self.config["sinol_expected_scores"])
+			results = self.perform_executions(compiled_commands, names, programs, self.args.expected_scores_report)
+			for prog in self.config["sinol_expected_scores"]:
+				for group, expected_result in self.config["sinol_expected_scores"][prog]["expected"].items():
```
tonowak, line 466: The `self.config["sinol_expected_scores"][prog]` shows up often and it makes the code a bit unreadable (because this long line takes most of my attention but it's just an implementation detail and is irrelevant)

src/sinol_make/commands/run/__init__.py
```diff
@@ -368,48 +377,98 @@ def print_view(output_file=None):
...
+				choice = input().lower()
+				if choice == "y" or choice == "yes" or choice == "":
+					with open(os.path.join(os.getcwd(), "config.yml"), "w") as f:
+						yaml.dump(new_config, f, default_flow_style=False)
+					print(util.color_green("Suggestions applied."))
+				else:
+					print("Suggestions not applied.")
+		else:
+			for prog in self.config["sinol_expected_scores"]:
+				program = os.path.join(self.EXECUTABLES_DIR, self.config["sinol_expected_scores"][prog]["validator"])
```
tonowak, line 461: I thought we ditched the idea of validators in `sinol_expected_scores`? What is the definition of a validator in this context and what should it do?

tonowak (commented):
Initial review -- today I don't have much time to look into it in more details.

src/sinol_make/commands/run/__init__.py
MasloMaslane: I copied this logic from the original code, but yes, it should work for group 0

src/sinol_make/commands/run/__init__.py
MasloMaslane: Let’s say that there are 3 groups of tests

```
sinol_expected_scores:
  very_cool_name1:
    program: abc1.cop
    points: 33
    expected: {1: OK, 2: WA, 3: ML}
  very_cool_name2:
    program: abc2.cop
    points: 67
    expected: {1: TL, 2: OK, 3: OK}
```

src/sinol_make/commands/run/__init__.py
MasloMaslane: Yes, I must have missed this while refactoring

src/sinol_make/commands/run/__init__.py
MasloMaslane: Yes,I think that this is a better approach

src/sinol_make/commands/run/__init__.py
```diff
@@ -368,48 +377,98 @@ def print_view(output_file=None):
...
+				print(util.bold(util.color_yellow('Suggestions will be applied.')))
+			else:
+				print(util.bold(util.color_yellow('Use flag --apply_suggestions to apply suggestions.')))
 
-				score_checksum += self.scores[group]
-			score_expected = self.config["subtasks"][subtask]["points"]
-			if score_checksum != score_expected:
-				print(util.bold(util.color_red('Subtask %s will grant %d points (expected %d).' % (subtask, score_checksum, score_expected))))
-				exit(1)
+		if suggestions:
```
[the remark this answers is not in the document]
MasloMaslane: Yes, I forgot about it

src/sinol_make/commands/run/__init__.py
MasloMaslane: Maybe I should add const `EXPECTED_SCORES = “sinol_expected_scores”` and use it?

src/sinol_make/commands/run/__init__.py
tonowak: We can make the following, which should be cleaner:

```
sinol_expected_scores:
  abc1.cpp:
    points: 33
    expected: {1: OK, 2: WA, 3: ML}
  abc2.cpp:
    points: 67
    expected: {1: TL, 2: OK, 3: OK}
```

src/sinol_make/commands/run/__init__.py
tonowak: > (bonus points for first printing all of than and only then exiting)

Minor suggestion: that probably can be achieved by first building the YAML that should be, and then comparing it with the existing YAML. Should be easy to handle any edge-cases with this method.

src/sinol_make/commands/run/__init__.py
tonowak: That won't make the code really shorter -- the string still is there, I still have to read it every three lines. I guess that first saving it to a variable and using this variable instead of `self.config["sinol_expected_scores"][prog]` could be better.

src/sinol_make/commands/run/__init__.py
tonowak: It's a good habit to read your own PR from the GitHub interface -- with this method, you probably could catch most of the comments I wrote here, which saves time for both of us.

MasloMaslane: OK, I’ll do it next time

src/sinol_make/commands/run/__init__.py
MasloMaslane: Sure

src/sinol_make/commands/run/__init__.py
MasloMaslane: I added `util.warning` and `util.error` which return colored strings.

src/sinol_make/commands/run/__init__.py
MasloMaslane: I removed the check for group 0.

src/sinol_make/commands/run/__init__.py
MasloMaslane: I changed it.

src/sinol_make/commands/run/__init__.py
MasloMaslane: It should now work as you described it.

src/sinol_make/commands/run/__init__.py
MasloMaslane: It should be more readable now.

tonowak:
Nice! The current code seems quite good and it's definitely close to being merged :100: 

Here are a few suggestions which would make it even better:

From what I understand (I'm not sure though (and that's a bad thing)), the solutions are being run in both `validate_expected_scores` and `run_programs`, which seems more than needed. From the name of the function `validate_expected_scores` I would expect it to only verify the YAML file. There is an easy way for checking whether a function needs a comment or a better name: try to describe in steps what it does and check whether this description can be shortened to the current name. (are the current function names good enough with this method?) We can also try to make `validate_expected_scores` testable with unit tests, by passing it all required data as function parameters and by not using IO (then it also would be possible to move it to another file -- the current file is getting pretty big).

I'm a bit worried about possible edge-cases in `validate_expected_scores`. For example:
- what happens when a solution gets removed? Is the tool checking that it indeed disappeared from the YAML file?
- what happens when a group gets removed or added?
- is the `--tests` flag currently handled correctly with the validation? (probably the most sense would be to not validate when this flag is present?)
- (...) and probably more which I haven't thought of yet.

I think that it's possible to write this code to make it more likely to work on edge-cases. Consider this approach:
- pass to `validate_expected_scores` a `results` dictionary generated by running the selected solutions on the tests and also pass the existing YAML file content,
- create a dictionary/list/whatever in the same format as the YAML field from the passed results,
- filter out non-relevant solutions from the original YAML field (e.g. by removing from it filenames from the set "all available solutions minus all solutions in `results`"),
- compare the content of both YAML fields (old vs. new). If it's the same, we are sure that nothing changed. If it's different, we can e.g. exchange it when `--apply-suggestions` is used or print both whole YAML fields  to the output for the user to visually compare what changed. (Though be wary of comparing lists with elements in different order),
- consider splitting it into even smaller functions.

What do you think? Can this be improved even further?

MasloMaslane:
In the current form `run_programs` is not needed, because `validate_expected_scores` does the same things and more. Currently `validate_expected_scores` runs all solutions and displays the table, but also checks that they have results as expected.

I'll rewrite validation of results so that it works better on edge cases.

MasloMaslane:
I rewrote the functions for validating expected scores and split it into smaller functions, it should work on edge cases.

src/sinol_make/commands/run/__init__.py
```diff
@@ -368,70 +381,176 @@ def print_view(output_file=None):
...
+				exit(1)
+			if result == "OK" and group != 0:
+				points += self.config["scores"][group]
+		return points
+
 
-		programs = []
-		for subtask in self.config["subtasks"]:
+	def validate_config(self, programs):
+		print("Validating config...")
```
tonowak, line 396: Is this function needed when we also validate all of its config in the `validate_expected_scores`?

tonowak (commented):
Here are a couple of suggestions that will make the code shorter, which will reduce the amount of tests you will have to write.

src/sinol_make/commands/run/__init__.py
```diff
@@ -133,11 +129,24 @@ def get_program_key(self, program):
...
 		if arg_problems is None:
-			all_programs = [program for program in os.listdir("prog/")
+			all_programs = [self.get_executable(program) for program in os.listdir("prog/")
 							if self.PROGRAMS_RE.match(program)]
 			return sorted(all_programs, key=self.get_program_key)
 		else:
-			return sorted(list(set(arg_problems)), key=self.get_program_key)
+			all_programs = []
+			for program in arg_problems:
+				if os.path.isfile(program):
```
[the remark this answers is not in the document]
MasloMaslane: This condition could be false if some solution was removed

src/sinol_make/util.py
```diff
@@ -112,3 +112,10 @@ def color_red(text): return "\033[91m{}\033[00m".format(text)
 def color_green(text): return "\033[92m{}\033[00m".format(text)
 def color_yellow(text): return "\033[93m{}\033[00m".format(text)
 def bold(text): return "\033[01m{}\033[00m".format(text)
+
+def info(text):
+	return bold(color_green(text))
+def warning(text):
+	return bold(color_yellow(text))
+def error(text):
+	return bold(color_red(text))
```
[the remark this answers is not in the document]
MasloMaslane: I agree, it will be more readable

src/sinol_make/commands/run/__init__.py
```diff
@@ -368,70 +381,176 @@ def print_view(output_file=None):
...
+				# Generate new config
+				expected_scores = self.config["sinol_expected_scores"]
+				for program in new_expected_scores.keys():
+					if program not in expected_scores: # if there is a new program found
+						expected_scores[program] = {
+							"expected": {},
+							"points": 0
+						}
+					for group, result in new_expected_scores[program]["expected"].items(): # Update expected results
+						expected_scores[program]["expected"][group] = result
```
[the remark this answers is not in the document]
MasloMaslane: I don’t know, it’s not required

src/sinol_make/commands/run/__init__.py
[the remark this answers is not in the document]
tonowak: > solution was removed

What do you mean exactly?

Is `arg_problem` the list of solutions sent through a CLI flag? If yes, it's more reasonable to make an error when a solution doesn't exist.

src/sinol_make/commands/run/__init__.py
```diff
@@ -368,70 +381,176 @@ def print_view(output_file=None):
...
+				new_expected_scores[program]["points"] = points
+
+			if self.args.apply_suggestions:
+				self.config["sinol_expected_scores"] = new_expected_scores
+				with open(os.path.join(os.getcwd(), "config.yml"), "w") as f:
+					yaml.dump(self.config, f, default_flow_style=False)
+				print(util.info("Saved suggested expected scores description."))
+			else:
+				print(util.warning("Use flag --apply_suggestions to apply suggestions."))
+				exit(1)
```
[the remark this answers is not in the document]
tonowak: > maybe add a message that there will be a lot of warnings printed since expected_scores is not set in config

Would that help the user in any way?

src/sinol_make/commands/run/__init__.py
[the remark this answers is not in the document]
tonowak: Then let's shorten the code by using already calculated variables.

src/sinol_make/commands/run/__init__.py
```diff
@@ -537,17 +653,24 @@ def run(self, args):
...
+
+			if self.args.apply_suggestions:
+				print(util.warning('Suggestions will be applied.'))
+			else:
+				print(util.warning('Use flag --apply_suggestions to apply suggestions.'))
+
+		programs = self.get_programs(self.args.programs)
+		self.validate_config(programs)
+		results = self.run_programs(programs)
+		self.validate_expected_scores(results, programs)
```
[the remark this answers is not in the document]
tonowak: Would that work correctly with the validation? What if the user runs on only one test? Will the tool inform the user that all solutions differ in all of the groups? E.g. that there was `expected: [OK, WA, TLE, WA]` but now it should be just e.g. `expected: [OK]` on all solutions?

src/sinol_make/commands/run/__init__.py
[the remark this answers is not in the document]
MasloMaslane: I removed this part of the code

src/sinol_make/commands/run/__init__.py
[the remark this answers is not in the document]
MasloMaslane: Now the code only compares the results for the tests and programs it was run on.

src/sinol_make/commands/run/__init__.py
MasloMaslane: I think we should remove this function, it doesn't add much

src/sinol_make/commands/run/__init__.py
```diff
@@ -35,14 +35,8 @@ def configure_subparser(self, subparser):
...
-		parser.add_argument('--validate_subtasks',
-							dest='validate_subtasks', action='store_true',
-							help='validate subtasks defined in config.yml')
-		parser.add_argument('--validate_programs',
-							dest='validate_programs', action='store_true',
-							help='validate programs in prog/ directory')
-		parser.add_argument('--subtask_report', type=str,
-							help='file to store report from subtask validation (in markdown)')
+		parser.add_argument('--expected_scores_report', type=str,
+							help='file to store report from expected scores validation (in markdown)')
```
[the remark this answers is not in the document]
MasloMaslane: Removed in 2c8aebd8dd21883f8eed88be5edb1a3b945bccf8

src/sinol_make/commands/run/__init__.py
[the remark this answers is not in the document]
MasloMaslane: Changed in 9030e667b7625301d9b2d4d5f4b50724a3ad49c2

src/sinol_make/commands/run/__init__.py
```diff
@@ -133,11 +129,24 @@ def get_program_key(self, program):
...
+				if os.path.isfile(program):
+					all_programs.append(self.get_executable(program))
+			return sorted(all_programs, key=self.get_program_key)
+
+
+	def get_solutions(self):
+		programs = self.get_programs(None)
+		solutions = []
+		for program in programs:
+			if not re.search("/inwer\./g", program) and not re.search("/ingen\./g", program) and not re.search("/chk\./g", program):
```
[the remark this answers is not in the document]
MasloMaslane: Changed in ec9e03f9652e1e21c0408bfcaeac2442483d2693

src/sinol_make/util.py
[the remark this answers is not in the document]
MasloMaslane: Added in 7077e1081d176d27da21fb28ae48d5194591be7b

src/sinol_make/commands/run/__init__.py
MasloMaslane: Removed function in 82c01ce5595419b14df61f5cd7ee7938837b4c00

src/sinol_make/commands/run/__init__.py
```diff
@@ -368,70 +379,153 @@ def print_view(output_file=None):
...
+			for program in self.config["sinol_expected_scores"].keys():
+				if program not in new_expected_scores.keys():
+					removed_programs.add(program)
+
+		# Only if sinol_make was run on all groups we should check for removed groups
+		if self.args.tests == None and "sinol_expected_scores" in self.config:
+			for program in self.config["sinol_expected_scores"].keys():
+				for group in self.config["sinol_expected_scores"][program]["expected"].keys():
+					if program in new_expected_scores and group not in new_expected_scores[program]["expected"].keys():
+						removed_groups.add(group)
```
tonowak, line 474: The removed programs here were created differently than the added programs. It seems inconsistent and there is more surface for bugs to appear. Was there a reason why they need to be created in other ways?

tonowak (commented):
Looks good! The code's logic is solid and we probably won't greatly change it further.

Try to split the code into even smaller functions. It should be possible -- there won't me that many arguments to the functions and it would increase the code readability. For example, I will be able to choose whether I want to read how the printing or how saving to the file is done, or whether I want to focus only on the code's logic.

src/sinol_make/commands/run/__init__.py
```diff
@@ -133,11 +127,25 @@ def get_program_key(self, program):
...
-			return sorted(list(set(arg_problems)), key=self.get_program_key)
+			all_programs = []
+			for program in arg_problems:
+				if not os.path.isfile(program):
+					raise Exception("Program %s does not exist" % program)
+				all_programs.append(self.get_executable(program))
+			return sorted(all_programs, key=self.get_program_key)
+
+
+	def get_solutions(self):
```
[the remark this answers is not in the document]
MasloMaslane: All names changed to solutions and executables a65daed364be1032ed87380272c25098df71e86e

src/sinol_make/commands/run/__init__.py
```diff
@@ -368,70 +379,153 @@ def print_view(output_file=None):
...
+						added_programs.add(program[0])
+				elif field[1] == "expected": # Groups were added
+					for group in change:
+						added_groups.add(group[0])
+			elif type == "change":
+				if field[1] == "expected": # Result for group changed
+					program = field[0]
+					group = field[2]
+					old_result = change[0]
+					group = change[1]
```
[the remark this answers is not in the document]
MasloMaslane: Typo fixed in a65daed364be1032ed87380272c25098df71e86e

src/sinol_make/commands/run/__init__.py
```diff
@@ -368,70 +379,153 @@ def print_view(output_file=None):
...
+				for program in expected_scores:
+					for group in removed_groups:
+						if group in expected_scores[program]["expected"]:
+							del expected_scores[program]["expected"][group]
+					expected_scores[program]["points"] = self.calculate_points(expected_scores[program]["expected"])
+
+				for program in new_expected_scores.keys():
+					if program in expected_scores:
+						for group, group in new_expected_scores[program]["expected"].items():
+							expected_scores[program]["expected"][group] = group
```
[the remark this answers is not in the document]
MasloMaslane: Typo fixed in a65daed364be1032ed87380272c25098df71e86e

src/sinol_make/commands/run/__init__.py
```diff
@@ -368,70 +379,153 @@ def print_view(output_file=None):
...
-
+		return self.perform_executions(compiled_commands, names, programs, self.args.program_report)
+
+
+	def print_expected_scores(self, expected_scores):
+		print("sinol_expected_scores:")
+		for program in expected_scores.keys():
+			print("  %s:" % program)
+			print("    points: %d" % expected_scores[program]["points"])
+			print("    expected: " + str(expected_scores[program]["expected"]))
```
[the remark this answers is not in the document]
MasloMaslane: Changed printing in 84141b413affa27b81118ff7cc16699f737a87bb

src/sinol_make/commands/run/__init__.py
```diff
@@ -368,70 +379,153 @@ def print_view(output_file=None):
...
+		print(util.bold("Expected scores from config:"))
+		self.print_expected_scores(expected_scores)
+		print(util.bold("\nExpected scores based on results:"))
+		self.print_expected_scores(new_expected_scores)
+
+		expected_scores_diff = dictdiffer.diff(expected_scores, new_expected_scores)
+		added_programs = set()
+		removed_programs = set()
+		added_groups = set()
+		removed_groups = set()
```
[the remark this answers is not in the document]
MasloMaslane: Changed in 954c7b27189063d343f5a1d075992c9f5a185e67

src/sinol_make/commands/run/__init__.py
MasloMaslane: Now removed programs and groups are checked with `dictdiffer`. 954c7b27189063d343f5a1d075992c9f5a185e67

MasloMaslane:
We should decide if after running on one solution and a group was removed, should this group be removed for all solutions or just the one that sinol-make was executed on.

src/sinol_make/commands/run/__init__.py
```diff
@@ -360,78 +367,169 @@ def print_view(output_file=None):
...
+					else:
+						expected_scores[solution] = new_expected_scores[solution]
+
+
+				self.config["sinol_expected_scores"] = expected_scores
+				with open(os.path.join(os.getcwd(), "config.yml"), "w") as f:
+					yaml.dump(self.config, f, default_flow_style=None)
+				print(util.info("Saved suggested expected scores description."))
+			else:
+				util.exit_with_error("Use flag --apply_suggestions to apply suggestions.")
```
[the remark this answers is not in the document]
MasloMaslane: We could certainly split the `validate_expected_scores` function into more smaller functions. I was planning to split the functions into different files in another PR after this one is merged.
I don't think that the `writer` is needed, because `pytest` has an easy way to capture printed text.

src/sinol_make/commands/run/__init__.py
```diff
@@ -537,17 +628,24 @@ def run(self, args):
...
-			self.validate_subtasks()
-		elif args.validate_programs:
-			self.run_programs()
-		else:
-			self.validate_subtasks()
-			self.run_programs()
+
+		if 'sinol_expected_scores' not in self.config.keys():
+			print(util.warning('Expected scores description not defined in config.yml. ' \
+				    	'The program will run all files on all tests and will print you the results. '))
```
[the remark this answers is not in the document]
MasloMaslane: Not really

src/sinol_make/commands/run/__init__.py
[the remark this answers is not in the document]
tonowak: Sounds good!

src/sinol_make/commands/run/__init__.py
```diff
@@ -360,78 +367,169 @@ def print_view(output_file=None):
...
+	def validate_expected_scores(self, results):
+		new_expected_scores = {} # Expected scores based on results
+
+		for solution in results.keys():
+			new_expected_scores[solution] = {
+				"expected": results[solution],
+				"points": self.calculate_points(results[solution])
+			}
+
+		expected_scores = {} # Expected scores from config with only solutions and groups that were run
```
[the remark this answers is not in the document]
MasloMaslane: Changed in b2dcd0237816e81201c8742f2bebb3849dd96ada

src/sinol_make/commands/run/__init__.py
```diff
@@ -360,78 +367,169 @@ def print_view(output_file=None):
...
+		run_solutions = results.keys() # Solutions that were run
```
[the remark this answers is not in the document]
MasloMaslane: Renamed in b2dcd0237816e81201c8742f2bebb3849dd96ada

src/sinol_make/commands/run/__init__.py
```diff
@@ -360,78 +367,169 @@ def print_view(output_file=None):
...
+		run_groups_set = set()
+		if self.args.tests == None and "sinol_expected_scores" in self.config: # If no groups were specified, use all groups from config
+			for solution in self.config["sinol_expected_scores"]:
+				for group in self.config["sinol_expected_scores"][solution]["expected"]:
+					run_groups_set.add(group)
+		else:
+			for solution in results.keys():
+				for group in results[solution].keys():
+					run_groups_set.add(group)
+		run_groups = list(run_groups_set) # Groups that were run
```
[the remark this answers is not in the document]
MasloMaslane: Changed in b2dcd0237816e81201c8742f2bebb3849dd96ada

src/sinol_make/commands/run/__init__.py
```diff
@@ -360,78 +367,169 @@ def print_view(output_file=None):
...
+
+		if "sinol_expected_scores" in self.config:
```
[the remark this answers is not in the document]
MasloMaslane: `self.config["sinol_expected_scores"]` moved into a variable in a198ea16fb1dd9cff5f9f737f4dc1d385c41c87b

src/sinol_make/commands/run/__init__.py
```diff
@@ -360,78 +367,169 @@ def print_view(output_file=None):
...
+					group = field[2]
+					old_result = change[0]
+					result = change[1]
+
+					print(util.warning("Solution %s passed group %d with status %s while it should pass with status %s." %
+										(solution, group, result, old_result)))
+
+		if len(added_solutions) > 0:
+			print(util.warning("Solutions were added: "), end='')
+			print(util.warning(", ".join(added_solutions)))
```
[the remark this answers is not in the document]
MasloMaslane: Added in 1ce798ddc4bb7a4ac0d6322874e3eb00ae71dab4

src/sinol_make/commands/run/__init__.py
```diff
@@ -360,78 +367,169 @@ def print_view(output_file=None):
...
+		if len(added_groups) > 0:
+			print(util.warning("Groups were added: "), end='')
+			print(util.warning(", ".join([str(group) for group in added_groups])))
+		if len(removed_groups) > 0:
+			print(util.warning("Groups were removed: "), end='')
+			print(util.warning(", ".join([str(group) for group in removed_groups])))
+
+		if expected_scores == new_expected_scores and \
+			len(added_solutions) == 0 and len(removed_solutions) == 0 and \
+			len(added_groups) == 0 and len(removed_groups) == 0:
```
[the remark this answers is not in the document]
MasloMaslane: Removed in f8768fbb81c462f3ff6eb103f31f4a0ddb1bf61f

src/sinol_make/commands/run/__init__.py
```diff
@@ -360,78 +367,169 @@ def print_view(output_file=None):
...
+							del expected_scores[solution]["expected"][group]
+					expected_scores[solution]["points"] = self.calculate_points(expected_scores[solution]["expected"])
+
+				for solution in new_expected_scores.keys():
+					if solution in expected_scores:
+						for group, result in new_expected_scores[solution]["expected"].items():
+							expected_scores[solution]["expected"][group] = result
+						expected_scores[solution]["points"] = self.calculate_points(expected_scores[solution]["expected"])
+					else:
+						expected_scores[solution] = new_expected_scores[solution]
```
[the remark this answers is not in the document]
MasloMaslane: Helper functions added in 37551d34afa4d4990e2708877663d557ce2b945f

src/sinol_make/commands/run/__init__.py
```diff
@@ -360,78 +367,190 @@ def print_view(output_file=None):
...
+
+		def warn_if_not_empty(set, message):
+			if len(set) > 0:
+				print(util.warning(message), end='')
+				print(util.warning(", ".join([str(x) for x in set])))
+
+		warn_if_not_empty(added_solutions, "Solutions were added: ")
+		warn_if_not_empty(removed_solutions, "Solutions were removed: ")
+		warn_if_not_empty(added_groups, "Groups were added: ")
+		warn_if_not_empty(removed_groups, "Groups were removed: ")
```
tonowak, line 508:
```suggestion
		def warn_if_not_empty(set, message):
			if len(set) > 0:
				print(util.warning(message + ": "), end='')
				print(util.warning(", ".join([str(x) for x in set])))

		warn_if_not_empty(added_solutions, "Solutions were added")
		warn_if_not_empty(removed_solutions, "Solutions were removed")
		warn_if_not_empty(added_groups, "Groups were added")
		warn_if_not_empty(removed_groups, "Groups were removed")
```

src/sinol_make/commands/run/__init__.py
```diff
@@ -360,78 +367,190 @@ def print_view(output_file=None):
...
+
+	def print_expected_scores_diff(self, validation_results):
+		added_solutions = validation_results["added_solutions"]
+		removed_solutions = validation_results["removed_solutions"]
+		added_groups = validation_results["added_groups"]
+		removed_groups = validation_results["removed_groups"]
+		changes = validation_results["changes"]
+		expected_scores = validation_results["expected_scores"]
+		new_expected_scores = validation_results["new_expected_scores"]
+		config_expected_scores = self.config["sinol_expected_scores"] if "sinol_expected_scores" in self.config else {}
```
tonowak, line 498: That looks like something that could be a `struct` (or whatever Python has). Having named fields reduces the risk of making a typo in the strings. Then, the function could e.g. take `validation_results` and then abbreviate it to `diff` or whatever, so that the names `diff.added_solutions` won't be long. That would also shorten the code thanks to not saving each dictionary value into a variable.

src/sinol_make/commands/run/__init__.py
```diff
@@ -360,78 +367,190 @@ def print_view(output_file=None):
...
+				print(util.warning(", ".join([str(x) for x in set])))
+
+		warn_if_not_empty(added_solutions, "Solutions were added: ")
+		warn_if_not_empty(removed_solutions, "Solutions were removed: ")
+		warn_if_not_empty(added_groups, "Groups were added: ")
+		warn_if_not_empty(removed_groups, "Groups were removed: ")
+
+		for change in changes:
+			print(util.warning("Solution %s passed group %d with status %s while it should pass with status %s." %
+										(change[0], change[1], change[2], change[3])))
```
tonowak, line 512: It's not clear which field contains what. I could move up to see a comment describing the fields, but a better solution would be to name those fields, through e.g. a struct.

src/sinol_make/commands/run/__init__.py
MasloMaslane: Changed in 189dd7f695d152db99a0ee7d6ddde26c4e1c1a45, 4e272b3503f388c825144586b1570e6b0b986c19 and 8e0951d73e667b2e353d0501fbfdcb2219a98c23.

src/sinol_make/commands/run/__init__.py
MasloMaslane: Changed in 189dd7f695d152db99a0ee7d6ddde26c4e1c1a45

src/sinol_make/commands/run/__init__.py
```diff
@@ -360,78 +367,184 @@ def print_view(output_file=None):
...
+		new_expected_scores = {} # Expected scores based on results
+
+		for solution in results.keys():
+			new_expected_scores[solution] = {
+				"expected": results[solution],
+				"points": self.calculate_points(results[solution])
+			}
+
+		config_expected_scores = self.config["sinol_expected_scores"] if "sinol_expected_scores" in self.config else {}
+		used_solutions = results.keys() # Solutions that were used
```
tonowak, line 413: Non-helpful comment. We can describe it in more detail or we can remove the comment.

src/sinol_make/commands/run/__init__.py
```diff
@@ -360,78 +367,184 @@ def print_view(output_file=None):
...
+			removed_groups,
+			changes,
+			expected_scores,
+			new_expected_scores
+		)
+
+
+	def print_expected_scores_diff(self, validation_results: ValidationResult):
+		diff = validation_results
+		config_expected_scores = self.config["sinol_expected_scores"] if "sinol_expected_scores" in self.config else {}
```
tonowak, line 492: I bet there is a Python function that retrieves a field in the dictionary when it exists and uses a default value otherwise. Using such a function here would make this line shorter.

tonowak (approved):
Seems good! The code is really short, seems like it handles all edge-cases and is readable -- hopefully that will both make it easier to test and easier to maintain :100: 

I left two small comments, after resolving them you can merge this PR :rocket:. Sorry for making so many comments :sweat_smile:

src/sinol_make/commands/run/__init__.py
MasloMaslane: I removed the comment 37a3d24b153913db62727496c4c3cba3d4dc75ef

src/sinol_make/commands/run/__init__.py
MasloMaslane: I found a nice way to do that f5f6eeb8c8faa875898c36d6d86290691f1ab8f3

PR #9, opened 2023-05-16
Introduce tests

MasloMaslane:
This pull request adds tests with `pytest`.


README.md
```diff
@@ -5,3 +5,9 @@ Currently in development and not yet ready to be used.
 ## Installing from source
 `pip3 install .`
 
+## Running tests
+1. Install `pytest` from `pip`: \
+```pip3 install pytest```
+2. Install `sinol-make`: \
+```pip3 install .```
```
tonowak, line 12: Can't we add `pytest` as a dependency, so it also gets downloaded during `pip3 install .`?

tonowak (commented):
Looks good! Basic test suite that probably will be extended in future PRs. It's both possible to write unit tests and integrations tests through IO. I like that.

Assuming we'll add more task packages for testing, it's reasonable to move `tests/abc` into a sub-directory of `tests`. After some tests we would probably end with mixed up test data and directories with source files, which makes navigation harder.

README.md
MasloMaslane: Sure, I changed it: 60ec487e1ce8d893f3fd8545666eae90e9b4da93

.gitignore
```diff
@@ -4,3 +4,4 @@ dist
 build
 .vscode
 __pycache__
+cache
```
tonowak, line 7: Does it work when `cache` is a subdirectory? I'm more familiar with e.g. `**/cache/`

tonowak (approved):

.gitignore
MasloMaslane: It works, I tried doing `/tests/packages/**/cache` to catch only cache folders in packages but for some reason it didn’t work

PR #10, opened 2023-05-19
Function for pretty saving the `config.yml`

MasloMaslane:
I couldn't find an easier and cleaner way of doing this. One of the advantages is that we have full control over what order and how the `config.yml` is saved.

### DIFF
--- a/src/sinol_make/util.py
+++ b/src/sinol_make/util.py
@@ -139,7 +139,8 @@ def save_config(config):
 					del config[field] # Same reason for deleting as above
 
 		if config != {}:
-			yaml.dump(config, config_file) # Save any custom fields
+			# All remaining non-considered fields are appended to the end of the file
+			yaml.dump(config, config_file)
 
 
 def color_red(text): return "\033[91m{}\033[00m".format(text)
--- a/src/sinol_make/util.py
+++ b/src/sinol_make/util.py
@@ -113,6 +113,8 @@ def save_config(config):
 	Function to save nicely formated config.yml
 	"""
 
+	# The order of the fields in the config.yml file for the sake of readability
+	# The fields that are not in this list will be appended to the end of the file
 	order = [
 		"title",
 		"memory_limit",
@@ -128,12 +130,12 @@ def save_config(config):
 	config = config.copy()
 	with open("config.yml", "w") as config_file:
 		for field in order:
-			if isinstance(field, dict):
+			if isinstance(field, dict): # If the field is a dict, it means that it has a custom property (for example default_flow_style)
 				if field["key"] in config:
 					yaml.dump({field["key"]: config[field["key"]]}, config_file, default_flow_style=field["default_flow_style"])
 					# The considered fields are deleted, thus `config` at then end will contain only custom fields written by the user
 					del config[field["key"]]
-			else:
+			else: # If the field is a string, it means that it doesn't have any custom properties, it's just a dict key
 				if field in config:
 					yaml.dump({field: config[field]}, config_file)
 					del config[field] # Same reason for deleting as above
--- a/src/sinol_make/util.py
+++ b/src/sinol_make/util.py
@@ -131,7 +131,8 @@ def save_config(config):
 			if isinstance(field, dict):
 				if field["key"] in config:
 					yaml.dump({field["key"]: config[field["key"]]}, config_file, default_flow_style=field["default_flow_style"])
-					del config[field["key"]] # Fields are deleted so that if there were any custom fields, they will be saved at the end of the file
+					# The considered fields are deleted, thus `config` at then end will contain only custom fields written by the user
+					del config[field["key"]]
 			else:
 				if field in config:
 					yaml.dump({field: config[field]}, config_file)
--- a/src/sinol_make/util.py
+++ b/src/sinol_make/util.py
@@ -1,4 +1,4 @@
-import glob, importlib, os, sys, subprocess, requests, tarfile
+import glob, importlib, os, sys, subprocess, requests, tarfile, yaml
 
 def get_commands():
 	"""
@@ -106,6 +106,38 @@ def get_oiejq_path():
 		return os.path.expanduser('~/.local/bin/oiejq')
 	else:
 		return None
+
+
+def save_config(config):
+	"""
+	Function to save nicely formated config.yml
+	"""
+
+	order = [
+		"title",
+		"memory_limit",
+		"time_limit",
+		"override_limits",
+		"scores",
+		{
+			"key": "sinol_expected_scores",
+			"default_flow_style": None
+		}
+	]
+	config_file = open("config.yml", "w")
+
+	for field in order:
+		if isinstance(field, dict):
+			if field["key"] in config:
+				yaml.dump({field["key"]: config[field["key"]]}, config_file, default_flow_style=field["default_flow_style"])
+				del config[field["key"]]
+		else:
+			if field in config:
+				yaml.dump({field: config[field]}, config_file)
+				del config[field]
+
+	yaml.dump(config, config_file)
+	config_file.close()
 
 
 def color_red(text): return "\033[91m{}\033[00m".format(text)
--- a/src/sinol_make/util.py
+++ b/src/sinol_make/util.py
@@ -146,6 +146,7 @@ def save_config(config):
 					del config[field] # Same reason for deleting as above.
 
 		if config != {}:
+			print(warning("Found unknown fields in config.yml: " + ", ".join([str(x) for x in config])))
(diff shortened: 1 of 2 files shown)


src/sinol_make/util.py
```diff
@@ -108,6 +108,38 @@ def check(path):
...
+			"default_flow_style": None
+		}
+	]
+	config_file = open("config.yml", "w")
+
+	for field in order:
+		if isinstance(field, dict):
+			if field["key"] in config:
+				yaml.dump({field["key"]: config[field["key"]]}, config_file, default_flow_style=field["default_flow_style"])
+				del config[field["key"]]
```
tonowak, line 133: Is there a reason for removing the field?

tonowak (commented):
Consider documenting it a bit more, e.g. by explaining the `if` statements. One of the issues with `if` statements that aren't obvious is that after time we will forget why they are there and we'll be scared of changing the code.

Other than that, it seems fine.

src/sinol_make/util.py
MasloMaslane: Yes, just before closing the file the rest of the config is saved. This could be helpful if there were some custom fields in the config.

tonowak: Then let's document this, this mechanism is not obvious.

src/sinol_make/util.py
```diff
@@ -108,6 +108,38 @@ def check(path):
...
+		else:
+			if field in config:
+				yaml.dump({field: config[field]}, config_file)
+				del config[field]
+
+	yaml.dump(config, config_file)
+	config_file.close()
```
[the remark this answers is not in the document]
MasloMaslane: Moved to `with` statement 1f5248277e1a2354f7a834e38a979a55312b28f7

src/sinol_make/util.py
MasloMaslane: Added description 59b6fe14207b5514a4592dd7fab08ff228abe24d

src/sinol_make/util.py
```diff
@@ -108,6 +108,39 @@ def check(path):
...
+		}
+	]
+
+	config = config.copy()
+	with open("config.yml", "w") as config_file:
+		for field in order:
+			if isinstance(field, dict):
+				if field["key"] in config:
+					yaml.dump({field["key"]: config[field["key"]]}, config_file, default_flow_style=field["default_flow_style"])
+					del config[field["key"]] # Fields are deleted so that if there were any custom fields, they will be saved at the end of the file
```
tonowak, line 134:
```suggestion
					yaml.dump({field["key"]: config[field["key"]]}, config_file, default_flow_style=field["default_flow_style"])
					# The considered fields are deleted, thus `config` at then end will contain only custom fields written by the user
					del config[field["key"]]
```

src/sinol_make/util.py
```diff
@@ -108,6 +108,39 @@ def check(path):
...
+			else:
+				if field in config:
+					yaml.dump({field: config[field]}, config_file)
+					del config[field] # Same reason for deleting as above
+
+		if config != {}:
+			yaml.dump(config, config_file) # Save any custom fields
```
tonowak, line 141:
```suggestion
			# All remaining non-considered fields are appended to the end of the file
			yaml.dump(config, config_file)
```

src/sinol_make/util.py
```diff
@@ -108,6 +108,39 @@ def check(path):
...
+			"key": "sinol_expected_scores",
+			"default_flow_style": None
+		}
+	]
+
+	config = config.copy()
+	with open("config.yml", "w") as config_file:
+		for field in order:
+			if isinstance(field, dict):
+				if field["key"] in config:
```
tonowak, line 132: Those `if`s are still not obvious to me. If you also think they are non-obvious, we can add comments explaining them.

MasloMaslane: I can add some comments here

MasloMaslane: 3fb7b53d5b20c2558451734177394e50e8774d77

src/sinol_make/util.py
```diff
@@ -108,6 +108,43 @@ def check(path):
...
+
+	config = config.copy()
+	with open("config.yml", "w") as config_file:
+		for field in order:
+			if isinstance(field, dict): # If the field is a dict, it means that it has a custom property (for example default_flow_style)
+				if field["key"] in config:
+					yaml.dump({field["key"]: config[field["key"]]}, config_file, default_flow_style=field["default_flow_style"])
+					# The considered fields are deleted, thus `config` at then end will contain only custom fields written by the user
+					del config[field["key"]]
+			else: # If the field is a string, it means that it doesn't have any custom properties, it's just a dict key
```
tonowak, line 138:
```suggestion
			else: # When the field is a string, it doesn't have any custom properties, so it's just a dict key
```

src/sinol_make/util.py
```diff
@@ -108,6 +108,43 @@ def check(path):
...
+		}
+	]
+
+	config = config.copy()
+	with open("config.yml", "w") as config_file:
+		for field in order:
+			if isinstance(field, dict): # If the field is a dict, it means that it has a custom property (for example default_flow_style)
+				if field["key"] in config:
+					yaml.dump({field["key"]: config[field["key"]]}, config_file, default_flow_style=field["default_flow_style"])
+					# The considered fields are deleted, thus `config` at then end will contain only custom fields written by the user
```
tonowak, line 136:
```suggestion
					# The considered fields are deleted, thus `config` at the end will contain only custom fields written by the user
```

src/sinol_make/util.py
```diff
@@ -108,6 +108,43 @@ def check(path):
 		return None
 
 
+def save_config(config):
+	"""
+	Function to save nicely formated config.yml
+	"""
+
+	# The order of the fields in the config.yml file for the sake of readability
+	# The fields that are not in this list will be appended to the end of the file
```
tonowak, line 117:
```suggestion
	# We add the fields in the config.yml in a particular order to make the config more readable
	# The fields that are not in this list will be appended to the end of the file
```

tonowak, line 117: Two such adjacent comments make it weird when there is no dot at the end of them -- it feels like something is missing and the sentences have a broken flow. What do you think about adding the dots and also adding dots in the other comments in the files to make it consistent?

MasloMaslane: I added dots to every comment in this function 24b28f1f13e26bc11b87306f79a0f31fd8f736a6

src/sinol_make/util.py
```diff
@@ -108,6 +108,43 @@ def check(path):
 		return None
 
 
+def save_config(config):
+	"""
+	Function to save nicely formated config.yml
+	"""
+
+	# We add the fields in the config.yml in a particular order to make the config more readable
+	# The fields that are not in this list will be appended to the end of the file
```
tonowak: Now that I think about it, what is a reasonable scenario where a user would have custom fields in the config? How would such fields be used? Maybe we should give a warning or an error in case of non-supported fields? I'm open for discussion.

tonowak:
Also, remember that any changes in the code style should also be applied to the other files in the project. I don't know how many comments there are, so we might want to make a separate PR that adds dots to all of them, or we can do it in this PR.

src/sinol_make/util.py
MasloMaslane: I've seen `oiejq_path` in some tasks (it was used by the `runner.py`), but other than that I don't recall seeing anything.

tonowak: I leave the decision up to you.

MasloMaslane:
I created an issue #11 for formating comments

src/sinol_make/util.py
MasloMaslane: I think we should leave this as is

tonowak: Sure, we can leave it. Are there any reasons for why you think it's the best choice?

MasloMaslane: I think that the `config.yml` should still be compatible with other scripts (`runner.py` for example). But I guess a warning would be nice.

MasloMaslane:
I also added some possible `config.yml` fields I missed: 4b1c1bf072ed58f81497ba1c3c9144e22ef6fd67

src/sinol_make/util.py
tonowak: Good point!

tonowak: Re-add me as a reviewer when you'll finish this PR (e.g. after adding the warning).

MasloMaslane: I added the warning: bab9b9b933e554db95061ea24205fbda93df1884

tonowak (approved):
:rocket:

PR #12, opened 2023-05-20
Add more tests for `run` command

MasloMaslane:
- [x] Add more unit tests for functions
 - [x] Add more specific integration tests

### DIFF
--- a/.github/workflows/Ubuntu.yaml
+++ b/.github/workflows/Ubuntu.yaml
@@ -1,0 +1,33 @@
+name: pytest-ubuntu
+run-name: Run pytest on Ubuntu
+on: [push]
+jobs:
+  pytest:
+    runs-on: self-hosted
+    container:
+      image: ubuntu:latest
+      volumes:
+        - /home/actions/oiejq:/github/home/.local/bin
+      env:
+        DEB_PYTHON_INSTALL_LAYOUT: deb
+      options:
+        --privileged
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Prepare system
+        run: |
+          apt update
+          export TZ=Europe/Warsaw
+          ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
+          sysctl kernel.perf_event_paranoid=-1
+      - name: Set up Python 3.11
+        run: |
+          apt install -y python3.11 python3.11-dev python3-pip
+          pip3 install -U setuptools
+      - name: Install dependencies
+        run: |
+          pip3 install .[tests]
+      - name: Run pytest
+        run: |
+          python3 -m pytest
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,5 @@ build
 .vscode
 __pycache__
 /tests/packages/**/cache
+/tests/packages/**/in/*.in
+/tests/packages/**/out/*.out
--- a/.gitignore
+++ b/.gitignore
@@ -5,5 +5,3 @@ build
 .vscode
 __pycache__
 /tests/packages/**/cache
-/tests/packages/**/in/*.in
-/tests/packages/**/out/*.out
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,7 @@ build
 .vscode
 __pycache__
 /tests/packages/**/cache
+
+# pytest-cov
+.coverage
+coverage.xml
--- a/src/sinol_make/commands/run/__init__.py
+++ b/src/sinol_make/commands/run/__init__.py
@@ -256,7 +256,12 @@ class Command(BaseCommand):
 		program_groups_scores = collections.defaultdict(dict)
 
 		def print_view(output_file=None):
-			stdout = sys.stdout # This is needed to restore stdout after printing to file. It must be a variable, because pytest can change default value of stdout.
+			def print_stream(*values, end='\n'):
+				if output_file is not None:
+					print(*values, end=end, file=output_file)
+				else:
+					print(*values, end=end)
+
 			if i != 0 and output_file is None:
 				# TODO: always display both tables
 				# if self.args.verbose:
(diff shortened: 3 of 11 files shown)


tonowak:
How should I treat the TODO list in the first comment of the PR when the PR is ready for review but the TODO list is not filled with checked out cells? Does it mean that you'll add more functionalities into this PR? In case it's a list of stuff to do in future PRs, it might be better to create an issue with this list, so that it don't become lost and we won't forget about it.

.gitignore
```diff
@@ -5,3 +5,5 @@ build
 .vscode
 __pycache__
 /tests/packages/**/cache
+/tests/packages/**/in/*.in
+/tests/packages/**/out/*.out
```
tonowak, line 9: Why do we ignore `in` and `out` files? Aren't there some situations where there might be useful in a test package?

src/sinol_make/commands/run/__init__.py
```diff
@@ -256,6 +256,7 @@ def perform_executions(self, compiled_commands, names, solutions, report_file):
 		program_groups_scores = collections.defaultdict(dict)
 
 		def print_view(output_file=None):
+			stdout = sys.stdout
```
tonowak, line 259: It seems you're using this variable only once in the code. Remember that each added variable increases the context a reader has to remember while trying to understand the whole code.

tests/commands/run/test_integration.py
```diff
@@ -1,21 +1,135 @@
...
+	command = Command()
+	command.run(args)
 
-		args = parser.parse_args(["run"])
-		command = Command()
+
+def test_no_expected_scores(capsys, create_package):
+	"""
+	Test with no sinol_expected_scores in config.yml.
+	Should run, but exit with exit code 1.
```
tonowak, line 25: In the code you're checking more than the exit code. We also can add what in the output you're checking.

Adding good comment messages greatly helps in understanding the code (even though they are tests). Testing codes tend to have a lot of IO and exceptions etc. which slows down reading the code.

tests/commands/run/test_integration.py
```diff
@@ -1,21 +1,135 @@
...
+
+	out = capsys.readouterr().out
+	assert "Solutions were added:" in out
+	assert "abc.cpp" in out
+
+
+def test_apply_suggestions(create_package):
+	"""
+	Test with no sinol_expected_scores in config.yml.
+	Verifies that suggestions are applied.
```
tonowak, line 53: We can extend this comment, currently it doesn't say much more than the function name.

tonowak (commented):
Seems good! Have you checked the code coverage with some tool?

I'm also curious how the whole tool's output looks like, because we'll have to fine-tune it for the users. But that can be done in a separate PR/Issue.

MasloMaslane:
The TODO was for me when I first created this PR as draft, but I forgot to update it, sorry.

MasloMaslane:
I never heard about any tools for checking code coverage. Could you send me an example?

src/sinol_make/commands/run/__init__.py
MasloMaslane: I know, but if there is an output file specified the `sys.stdout` is changed to that file. And I can't simply set this value again to `sys.stdout` and the end of the function because then the `pytest`'s output capture doesn't work.

.github/workflows/Ubuntu.yaml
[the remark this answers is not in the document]
MasloMaslane: No, I merged #13 into this branch

.gitignore
MasloMaslane: I added this because I was hand testing in `abc` package and created the in/out files. And since I was lazy, instead of deleting these files I added them to .gitignore. I'll delete them now.

tests/commands/run/test_integration.py
MasloMaslane: I added better comments: 32873ce9a2f8f9c627efa0c1dc127260b0c08569

tests/commands/run/test_integration.py
MasloMaslane: I changed some comments: 32873ce9a2f8f9c627efa0c1dc127260b0c08569

.gitignore
MasloMaslane: Removed from .gitignore: 2961047328dd96be34def8f089f30e9738dcaa7f

src/sinol_make/commands/run/__init__.py
tonowak: I'm not sure I understand what you mean. Nevertheless it sounds like something that can be documented in a comment.

MasloMaslane: Is it understandable now? 866ba0d2cf1a2b5e6e92cba382d324e548bde0fe

tonowak: Do I understand correctly that `pytest` can change any variables named `stdout` anywhere in the code for testing purposes, and you're using this mechanic with a default value `sys.stdout`? I'm still not sure how does it work and why it must be a variable.

MasloMaslane: The default value of `sys.stdout` is `sys.__stdout__` which is a normal terminal stdout. This function sometime changes `sys.stdout` to a file. Previously, the function at the end returned `sys.stdout` to `sys.__stdout__`. But now the problem is that `pytest` changes `sys.stdout` to some custom stream to capture the output. So the easiest way to return `sys.stdout` to a correct stream at the end of the function is to have a variable that remembers what the stdout was.

tonowak: How does `sys.stdout` change? That is, why it is "not correct" at the end?

MasloMaslane: This function can set `sys.stdout` to a file in which the executions' table will be saved. And at the end we want to return `sys.stdout` to the previous state.

tonowak: `sys.stdout = open(output_file, 'w')`. OK, I see now, thanks.

I leave it up to you to change the comment to make it more clear.

MasloMaslane: What do you think about a comment like that:
`# sys.stdout is stored in a variable for an easy way of restoring it in case sys.stdout is changed to a file in this function`

tonowak: It's often better to start with the reason:
```
This function changes `sys.stdout` for easier printing to a file.
Because of that, we store `sys.stdout` to a variable for reverting that at the end of the function.
```

MasloMaslane: Okay, thanks

tonowak: We can also consider using a `print` alternative that accepts a stream (file or stout) as an argument.

MasloMaslane: There is an option in `print` to specify output stream, but then everywhere in this function we'd have to remember to specify it everywhere in this function

tonowak (approved):

PR #13, opened 2023-05-21
GitHub workflow

MasloMaslane:
After a lot of debugging workflow finally works. Currently tests are only ran on latest ubuntu, but adding other systems that can run in `docker` container should be easy.

### DIFF
--- a/.github/workflow/pytest.yaml
+++ b/.github/workflow/pytest.yaml
@@ -1,6 +1,11 @@
 name: pytest
 run-name: Run pytest
-on: [push, pull_request]
+on:
+  push:
+    branches: '*'
+  pull_request:
+    branches: '*'
+
 jobs:
   pytest:
     runs-on: self-hosted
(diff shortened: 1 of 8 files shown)


MasloMaslane:
What we could do is on push run tests on only one system (e. g. ubuntu) and when merging we can run tests on all systems. Here is how we could do this: https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue#triggering-merge-group-checks-with-github-actions

tonowak:
Is there a reason why we should limit ourselves to only one system for pushes? I bet we could run all systems without any problems. It shouldn't slow down the GitHub jobs too much because of concurrency and it might find some bugs in the tool sooner.

tonowak (approved):
Probably we'll want to extend the workflow's functionality, but currently it's a very good start! :rocket:

MasloMaslane:
Oh, I didn't know that GitHub jobs are concurrent. If so, we can run jobs on every push

PR #14, opened 2023-05-22
Add version flag

MasloMaslane:
In preparation for release, I added --version flag.
For now the version in `pyproject.toml` is 0.0.1, but I think we should release this as 1.0.0

### DIFF
--- a/src/sinol_make/__init__.py
+++ b/src/sinol_make/__init__.py
@@ -6,6 +6,7 @@ def configure_parsers():
 		prog='sinol-make',
 		description='Tool for creating and testing sio2 tasks',
 	)
+	parser.add_argument("-v", "--version", action="version", version="%(prog)s 0.0.1")
 	subparsers = parser.add_subparsers(
 		title='commands',
 		description='sinol-make commands',
--- a/src/sinol_make/__init__.py
+++ b/src/sinol_make/__init__.py
@@ -1,4 +1,4 @@
-import argparse
+import argparse, pkg_resources
 from sinol_make import util
 
 def configure_parsers():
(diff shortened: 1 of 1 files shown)


tonowak:
Maybe there is a way to get the version from project.toml? I bet someday we'll forget to change the version in the source code.

tonowak:
> For now the version in pyproject.toml is 0.0.1, but I think we should release this as 1.0.0

Sure, it doesn't really matter. Can I assume that you're following semantic versioning?

src/sinol_make/__init__.py
```diff
@@ -1,12 +1,12 @@
...
+import argparse, pkg_resources
 from sinol_make import util
 
 def configure_parsers():
 	parser = argparse.ArgumentParser(
 		prog='sinol-make',
 		description='Tool for creating and testing sio2 tasks',
 	)
-	parser.add_argument("-v", "--version", action="version", version="%(prog)s 0.0.1")
+	parser.add_argument("-v", "--version", action="version", version="%(prog)s " + pkg_resources.get_distribution('sinol_make').version)
```
tonowak, line 9: Is the `s` in `%(prog)s` on purpose or is it a typo?

MasloMaslane: No, it's there to specify that the `prog` should be converted to string

tonowak: Ok, I see. So the output will be `sinol-make: 0.0.1`?

MasloMaslane:
I added automatic detection of the version from package info b6857cac3bf6f7a2bdafda9e46a98fffe3b7fd87

tonowak (approved):

MasloMaslane:
And I'm planning to follow the semantic versioning

PR #15, opened 2023-05-22
Add release job

MasloMaslane:
I added release job that runs on GitHub release. I think that it should work.


MasloMaslane:
I can test it on some other repository

tonowak:
> There is also the option to create a dummy package in the pip's testing repository.

https://packaging.python.org/en/latest/guides/using-testpypi/

MasloMaslane:
OK, I'll test it tomorrow

MasloMaslane:
The workflow works. You can install sinol-make from test pypi: `pip3 install --index-url https://test.pypi.org/legacy --extra-index-url https://pypi.org/simple/ sinol-make`

tonowak:
Great! Then let's change it to the proper pypi repo and merge it :)

PR #16, opened 2023-05-23
Remove deprecated package `pkg_resources`

MasloMaslane:
Package `pkg_resources` is deprecated. I found another way of having current version in one place.


tonowak (approved):
That's... interesting.
If that's the best way of doing it, you can merge it.

MasloMaslane:
I didn't find any other working way of doing this.

PR #19, opened 2023-06-01
Add option to measure time and memory consumption with `time`

MasloMaslane:
I added tests in another PR #21


src/sinol_make/commands/run/__init__.py
```diff
@@ -202,41 +204,82 @@ def execute(self, execution):
...
+			return result
+		elif self.args.time_tool == 'time':
+			if sys.platform == 'darwin':
+				command = 'gtimeout -k %ds %ds gtime -f "%%U\\n%%M\\n%%x" -o %s %s <%s >%s' \
+					% (hard_time_limit_in_s, hard_time_limit_in_s,
+						result_file, executable, test, output_file)
+			elif sys.platform == 'linux':
+				command = 'timeout -k %ds %ds time -f "%%U\\n%%M\\n%%x" -o %s %s <%s >%s' \
+					% ( hard_time_limit_in_s, hard_time_limit_in_s,
+						result_file, executable, test, output_file)
```
tonowak, line 253: That's some code duplication which can be shortened.

src/sinol_make/commands/run/__init__.py
```diff
@@ -202,41 +204,82 @@ def execute(self, execution):
...
-			result["Status"] = "TL"
-		elif "Time" in result.keys() and result["Time"] > time_limit:
-			result["Status"] = "TL"
-		elif "Memory" in result.keys() and result["Memory"] > memory_limit:
-			result["Status"] = "ML"
-		elif "Status" not in result.keys():
-			result["Status"] = "RE"
-		elif result["Status"] == "OK":
-			if result["Time"] > time_limit:
+		if self.args.time_tool == 'oiejq':
```
tonowak, line 207: Maybe it would be better to split the time vs oiejq `if` to separate documented functions? Currently it's a bit long and it's hard to keep track of what is being changed and what are the assumptions of those ifs.

src/sinol_make/commands/run/__init__.py
```diff
@@ -202,41 +204,82 @@ def execute(self, execution):
...
+
+			timeout_exit_code = os.system(command)
+
+			result = {}
+			lines = open(result_file).readlines()
+			program_exit_code = None
+			if len(lines) == 3:
+				result["Time"] = round(float(lines[0].strip()) * 1000)
+				result["Memory"] = int(lines[1].strip())
+				program_exit_code = int(lines[2].strip())
```
tonowak, line 265: When doing such hacky parsing it's good to write a comment with an example output, so that I can see whether the code makes sense on an example. Also, the `time` output might differ for different operating systems or even different shells in the same operating system, so it might be beneficial to assert how the output looks like, or maybe both parse and assert at once.

src/sinol_make/commands/run/__init__.py
```diff
@@ -202,41 +204,82 @@ def execute(self, execution):
...
+			elif timeout_exit_code != 0:
+				result["Status"] = "TL"
+			elif result["Time"] > time_limit:
 				result["Status"] = "TL"
 			elif result["Memory"] > memory_limit:
 				result["Status"] = "ML"
-			elif os.system("diff -q -Z %s %s >/dev/null"
-						% (output_file, self.get_output_file(test))):
+			elif os.system("diff -q -Z %s %s >/dev/null" % (output_file, self.get_output_file(test))):
 				result["Status"] = "WA"
```
tonowak, line 278: It's hard to keep track of the `result[X]` data while reviewing, we can consider doing a documented struct. I leave the decision up to you.

tests/packages/abc/prog/abc4.cpp
```diff
@@ -2,6 +2,10 @@
 
 using namespace std;
 
+int rnd(int a, int b) {
+	return a + rand() % (b - a + 1);
+}
+
```
tonowak, line 8: This function here is not needed, right?

src/sinol_make/commands/run/__init__.py
MasloMaslane: Splitted parsing results into different functions: 09ad06b71c4941db2439cfaabd9116b51c9f2873

src/sinol_make/commands/run/__init__.py
MasloMaslane: Some comments about output of time command: fbbc71e26f9a547237dbe255700ba132d5738eb1. I'm using time's -f flag which specifies the format.

src/sinol_make/commands/run/__init__.py
MasloMaslane: Struct for results: 5a10ad090d88443c6d28df93fe1bb9ca699eb389

tests/packages/abc/prog/abc4.cpp
MasloMaslane: Yes, I deleted it: 486727169eab6172e00a446bf84c28e1173c8ee1

README.md
tonowak, line 1: Let's move those installation steps to the PR with the README.
Additionally, let's deliberately not write installation steps for `time` on Linux-based systems -- I want to as much as possible force the users to use `oiejq`.

src/sinol_make/commands/run/__init__.py
```diff
@@ -38,6 +38,8 @@ def configure_subparser(self, subparser):
 							help='hide memory usage in report')
 		parser.add_argument('--program_report', type=str,
 							help='file to store report from program executions (in markdown)')
+		parser.add_argument('--time_tool', choices=['oiejq', 'time'], default='oiejq',
+		      				help='tool to measure time and memory usage (default: oiejq)')
```
tonowak, line 42: Is it also default for macOS, for which oiejq doesn't work?

src/sinol_make/commands/run/__init__.py
```diff
@@ -194,50 +196,114 @@ def compile(self, solution):
...
+		timeout_exit_code = os.system(command)
+
+		result = ExecutionResult(None, None, None)
+		lines = open(result_file).readlines()
+		program_exit_code = None
+		if len(lines) == 3:
+			"""
+			If programs runs successfully, the output looks like this:
+			 - first line is CPU time in seconds
+			 - second line is memory in Kbytes
```
tonowak, line 246:
```suggestion
			 - second line is memory in KB
```

src/sinol_make/commands/run/__init__.py
```diff
@@ -194,50 +196,114 @@ def compile(self, solution):
...
+			result.Status = "WA"
 		else:
-			result["Status"] = result["Status"][:2]
+			result.Status = "OK"
+
 		return result
 
+
+	def execute(self, execution):
+		(name, executable, test, time_limit, memory_limit, timetool_path) = execution
```
tonowak, line 278: The function name and argument name are quite similar and thus I'm not sure what the function exactly does based on its signature. It's also quite similar to the function name `perform_executions`, which also confused me. Let's find more intuitive names to the functions/arguments.

src/sinol_make/commands/run/__init__.py
```diff
@@ -194,50 +196,114 @@ def compile(self, solution):
...
+		output_file = os.path.join(self.EXECUTIONS_DIR, name,
+								self.extract_test_no(test)+".out")
+		result_file = os.path.join(self.EXECUTIONS_DIR, name,
+								self.extract_test_no(test)+".res")
+		hard_time_limit_in_s = math.ceil(2*time_limit / 1000.0)
```
tonowak, line 283:
```suggestion
								self.extract_test_no(test) + ".out")
		result_file = os.path.join(self.EXECUTIONS_DIR, name,
								self.extract_test_no(test) + ".res")
		hard_time_limit_in_s = math.ceil(2 * time_limit / 1000.0)
```
For consistency. Also, I wonder, is there an elegant way to create both `output_file` and `result_file` variables without code duplication? It's not important though, so feel free to skip this question.

src/sinol_make/commands/run/structs.py
```diff
@@ -17,3 +17,13 @@ class ValidationResult:
...
 	new_expected_scores: dict
+
+@dataclass
+class ExecutionResult:
+	# Result status of execution. Can be one of:
+	# "OK", "WA", "TL", "ML", "RE", "CE"
+	Status: str
+	# Time in milliseconds
+	Time: float
+	# Memory in Kbytes
```
tonowak, line 28:
```suggestion
	# Memory in KB
```

src/sinol_make/commands/run/__init__.py
MasloMaslane: Yes, I will change it

src/sinol_make/commands/run/structs.py
MasloMaslane: I don't think it should be, but there is 
`math.ceil(memory_limit)` 
and
`self.memory_limit = args.ml * 1024.0 if args.ml is not None else config_memory_limit`
for some reason

tonowak: Maybe the author also wasn't sure whether it's a float or an int. The memory is getting retrieved from oiejq, right? Maybe looking into sio2jail might find the answer.

src/sinol_make/commands/run/__init__.py
MasloMaslane: Changed 3be1569831372fb44a4161a43df1331e561cd0de

src/sinol_make/commands/run/structs.py
MasloMaslane: I read a bit and I think that memory is always an int 739360d47a75e70a998138588eb032b1b63dd1f8

src/sinol_make/commands/run/__init__.py
MasloMaslane: I changed names of the fuctions 7215e30ccd3d94fe1e84eb4080a27ef9102abfaa

src/sinol_make/commands/run/__init__.py
```diff
@@ -38,6 +41,8 @@ def configure_subparser(self, subparser):
 							help='hide memory usage in report')
 		parser.add_argument('--program_report', type=str,
 							help='file to store report from program executions (in markdown)')
+		parser.add_argument('--time_tool', choices=['oiejq', 'time'], default=default_timetool,
+		      				help='tool to measure time and memory usage (default: oiejq)')
```
tonowak, line 45:
```suggestion
		      				help='tool to measure time and memory usage (default when possible: oiejq)')
```

src/sinol_make/commands/run/__init__.py
```diff
@@ -194,63 +199,134 @@ def compile(self, solution):
...
+		else:
+			result.Status = "OK"
+
+		return result
+
+
+	def run_execution(self, execution):
+		"""
+		Run an execution and return the result as ExecutionResult object.
+		"""
```
tonowak, line 283: Still, what is the `execution` argument? The function name suggests that it's somehow "running" the argument, but the first line of code suggest that this argument contains the results of a previously-done execution, but then, what does this function do?

src/sinol_make/commands/run/structs.py
```diff
@@ -17,3 +17,13 @@ class ValidationResult:
...
+
+@dataclass
+class ExecutionResult:
+	# Result status of execution. Can be one of:
+	# "OK", "WA", "TL", "ML", "RE", "CE"
+	Status: str
+	# Time in milliseconds
+	Time: int
+	# Memory in KB
+	Memory: float
```
tonowak, line 29: It's the memory that should be in int, right? (but maybe the time also, which would make sense when it's in milliseconds)

tonowak (commented):
Hopefully last comments and suggestions:

src/sinol_make/commands/run/__init__.py
MasloMaslane: This function runs a solution on the specified test and returns status code and time and memory used. I don't really have an idea how to call this function

src/sinol_make/commands/run/structs.py
tonowak: Usually it's enough to read the code in the GitHub interface before re-adding someone as a reviewer. The reviewer often just re-reads the whole code and if you're doing it first, it allows to skip a lot of going back-and-forth with the review.

src/sinol_make/commands/run/__init__.py
MasloMaslane: Maybe `def run_solution_on_test(test_info)` ?

tonowak: I would say that `run_solution(single_test_info)` is enough -- it avoids a repetition and underlines that it's a single test (though I'm not sure if it's actually the case -- I'm making the assumption based on the name `test_info` instead of `tests_info`). Though then, what kind of solution does it run, e.g. how does it know which solution to run? I guess it's also in the `test_info`, but that means that the name is a bit misleading.

MasloMaslane: Then maybe `execution_info` or `run_info`? But then again, it uses the unhelpful `execution` word.

tonowak: `data_for_execution` implies that it contains multiple details about the not-yet-done execution that will be performed in this function, which I think is a quite readable name. To even more improve the readability, we can consider creating a class that will contain fields for `data_for_execution`, and then passing this class as an argument.

MasloMaslane: I added the struct for execution data 4eb4acbc0273b501d174e5eed6bf132288a26a0b

src/sinol_make/commands/run/__init__.py
```diff
@@ -615,21 +691,29 @@ def run(self, args):
...
+					util.exit_with_error('Invalid oiejq path.')
+				self.timetool_path = args.oiejq_path
+			else:
+				self.timetool_path = util.get_oiejq_path()
+			if self.timetool_path is None:
+				util.exit_with_error('oiejq is not installed.')
+		elif args.time_tool == 'time':
+			if sys.platform == 'win32' or sys.platform == 'cygwin':
+				util.exit_with_error('Measuring with `time` is not supported on Windows.')
+			self.timetool_path = None
```
tonowak, line 708: I feel like those `if`s would be better placed somewhere where the command-line arguments are handled. This logic doesn't really suit to a function named `print_expected_scores`, as it doesn't have to do with printing.

tonowak (approved):
Looks good! I left one comment, you can merge after resolving it (or after ignoring it, it's not an important comment).

src/sinol_make/commands/run/__init__.py
MasloMaslane: Changed ba6e23795437d0426d93d39f52f4ce701eb219c4

tonowak (approved):
:100:

PR #20, opened 2023-06-01
Allow only `gcc` in versions 9 and 10 on MacOS

MasloMaslane:
MacOS has new version of g++ and gcc compilers. Now the tool will check for their existence, but I wonder if there is a better way of doing this.

### DIFF
--- a/src/sinol_make/commands/run/__init__.py
+++ b/src/sinol_make/commands/run/__init__.py
@@ -41,9 +41,9 @@ class Command(BaseCommand):
 		parser.add_argument('--oiejq_path', type=str,
 		      				help='path to oiejq executable (default: `~/.local/bin/oiejq`)')
 		parser.add_argument('--c_compiler_path', type=str, default=compiler.get_c_compiler_path(),
-		    				help='C compiler to use (default for Linux and Windows: gcc, default for Mac: gcc-{9-12})')
+		    				help='C compiler to use (default for Linux and Windows: gcc, default for Mac: gcc-{9,10})')
 		parser.add_argument('--cpp_compiler_path', type=str, default=compiler.get_cpp_compiler_path(),
-		    				help='C++ compiler to use (default for Linux and Windows: g++, default for Mac: gcc-{9-12})')
+		    				help='C++ compiler to use (default for Linux and Windows: g++, default for Mac: g++-{9,10})')
 		parser.add_argument('--python_interpreter_path', type=str, default=compiler.get_python_interpreter_path(),
 		    				help='Python interpreter to use (default: python3)')
 		parser.add_argument('--java_compiler_path', type=str, default=compiler.get_java_compiler_path(),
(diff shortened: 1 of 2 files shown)


tonowak:
Maybe https://stackoverflow.com/a/71128673 ?

But I see another problem -- it seems that it's possible to install the `gcc` on MacOS in another way than `brew` and then it's available through simply `gcc` instead of e.g. `gcc-11`.

MasloMaslane:
MacOS comes with `g++` and `gcc` installed, but those are for clang. I don't think it's possible to have `g++` and `gcc` installed as GNU compiler, unless someone links `g++-whatever` to `g++`

tonowak:
Is it possible to install a specific `g++` version alongside other existing versions on the MacOS? I'm thinking whether it's reasonable to require from the user the exact same version of `g++` that is used in `sio2`.

MasloMaslane:
Yes, it's possible

tonowak:
Then let's go with this. Hopefully thanks to that we won't have to maintain the supported `gcc` versions and maybe it would help the users through forcing them to use the same version of the compiler.

MasloMaslane:
I think that version 10 of `gcc` is used in oioioi, but this version doesn't support MacOS with ARM. Maybe we should check for gcc-10 and gcc-11?

tonowak:
Sounds good.

src/sinol_make/commands/run/__init__.py
```diff
@@ -41,9 +41,9 @@ def configure_subparser(self, subparser):
 		parser.add_argument('--oiejq_path', type=str,
 		      				help='path to oiejq executable (default: `~/.local/bin/oiejq`)')
 		parser.add_argument('--c_compiler_path', type=str, default=compiler.get_c_compiler_path(),
-		    				help='C compiler to use (default for Linux and Windows: gcc, default for Mac: gcc-{9-12})')
+		    				help='C compiler to use (default for Linux and Windows: gcc, default for Mac: gcc-{9,10})')
 		parser.add_argument('--cpp_compiler_path', type=str, default=compiler.get_cpp_compiler_path(),
-		    				help='C++ compiler to use (default for Linux and Windows: g++, default for Mac: gcc-{9-12})')
+		    				help='C++ compiler to use (default for Linux and Windows: g++, default for Mac: g++-{9,10})')
```
tonowak, line 46:
```suggestion
		    				help='C++ compiler to use (default for Linux and Windows: g++, default for Mac: g++-9 or g++-10)')
```

src/sinol_make/helpers/compiler.py
```diff
@@ -25,7 +25,7 @@ def get_c_compiler_path():
 		else:
 			return 'gcc'
 	elif sys.platform == 'darwin':
-		for i in range(9, 13):
+		for i in range(9, 11):
```
tonowak, line 28: `for i in [10, 9]:` is both more readable and first tries the newer compiler version (because the version 10 is the preferred one, do I remember correctly?).

MasloMaslane: Version 9 is currently used in oioioi, but it isn't available on MacOS with ARM.

tonowak: Ok, then let's leave it with the order `9, 10`.

tonowak, line 28: that one is `range`, but the below one is `[]`

MasloMaslane: Sorry, I forgot about it

tonowak (approved):

PR #21, opened 2023-06-01
Tests for measuring with `time`

MasloMaslane:


MasloMaslane:
I don't think that any new tests are needed for measuring with `time`.

tonowak (approved):
Nice! That's a great way of testing it without writing duplicate code, I like it!

Don't merge this PR until the other branch will be approved. Otherwise the files from this PR will need to be reviewed/approved also in the other PR.

PR #22, opened 2023-06-04
Initial README

tonowak:


README.md
```diff
@@ -1,11 +1,48 @@
-# sinol-make
-CLI tool for creating sio2 task packages. \
-Currently in development and not yet ready to be used.
+# sinol-make ![](https://avatars.githubusercontent.com/u/2264918?s=200&v=4)
```
tonowak, line 1: The width needs to be set or the whole image needs to be in another place.

MasloMaslane (commented):
Looks good to me, but we should either describe how to properly install `oiejq` or we should download it automatically (and maybe mention somewhere where it's installed).

tonowak:
> or we should download it automatically

Do you want to try implementing it?

MasloMaslane:
The function is already there, it's just not used anywhere

MasloMaslane (approved):

tonowak:
Oh. Seeing that it's implemented and not used, is there anything blocking it from being a part of the CLI?

README.md
```diff
@@ -1,11 +1,48 @@
...
+they lack some built-in mechanisms for verifying packages and finding mistakes
+before uploading the package to the judge system.
+
+### Installation
+
+It's possible to directly install [sinol-make](https://pypi.org/project/sinol-make/)
+through Python's package manager pip, which usually is installed alongside Python:
+
+`pip3 install sinol-make`
+
```
tonowak, line 34:
```suggestion

`pip3 install sinol-make`

We support `macOS` through the usage of `time` instead of `oiejq`.
As using `oiejq` is highly recommended, we suggest using a Linux-based
operating system whenever possible.

```

README.md
```diff
@@ -1,11 +1,48 @@
...
+- [Reporting bugs and contributing code](#reporting-bugs-and-contributing-code)
+
+### Why?
+
+The purpose of the tool is to make it easier to create good problem packages
+for official competitions, which requires collaboration with other people
+and using a multitude of "good practices" recommendations.
+While there are several excellent CLI tools for creating tests and solutions,
+they lack some built-in mechanisms for verifying packages and finding mistakes
+before uploading the package to the judge system.
```
tonowak, line 26:
```suggestion
before uploading the package to the judge system.
As sinol-make was created specifically for the sio2 platform,
by default it downloads and uses sio2's deterministic mechanism of measuring
a solution's runtime, called `oiejq`.
```

tonowak (commented):
Can you confirm @MasloMaslane ?

README.md
MasloMaslane: `oiejq` doesn't also work on WSL on Windows, so we should add it here

README.md
MasloMaslane: Yeah, I like that

README.md
tonowak: I've updated the suggestion. If you think it's good, you can clik on the "commit"

tonowak:
I wonder, maybe Python has some packages/built-in functions that can measure time? That would reduce the need to download a shell command (e.g. `time`) and would make the installation steps easier.

MasloMaslane:
I don’t think there is a better way to measure time. But I’ll check how `sioworkers` do it, maybe there is some other interesting way

MasloMaslane:
`sioworkers` also use time for measuring, so I think we shouldn't change it. Especially that when there will be a dedicated library for interacting with sinol packages, it will run programs in the same way as `sioworkers`do now.

tonowak:
Sounds good, let's leave it as it is.

README.md
```diff
@@ -1,11 +1,62 @@
...
+As sinol-make was created specifically for the sio2 problem packages,
+by default it downloads and uses sio2's deterministic mechanism of measuring
```
MasloMaslane, line 29: Technically, oiejq is just a wrapper for sio2jail, and sio2 doesn’t use oiejq

README.md
```diff
@@ -1,11 +1,62 @@
...
+solutions' runtime, called `oiejq`.
```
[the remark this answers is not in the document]
tonowak:
```suggestion
solutions' runtime, wrapped as `oiejq`.
```
Would that be closer to the truth? It might now be harder to understand the sentence though.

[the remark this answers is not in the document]
MasloMaslane: Yeah, I think that it's harder to understand now. I think we should leave it as it is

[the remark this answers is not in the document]
tonowak: Then I'll leave it as it is. Is it ready to be merged, or do you have some more comments/suggestions?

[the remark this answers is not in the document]
MasloMaslane: I think it's ready to merge

PR #23, opened 2023-06-14
Install `oiejq` automatically

MasloMaslane:


tonowak (commented):
How about testing this functionality? If something here doesn't work as intended, the user won't have a functioning `sinol-make`.

src/sinol_make/__init__.py
```diff
@@ -30,6 +32,20 @@ def main():
...
 	for command in commands:
 		if command.get_name() == args.command:
+			if sys.platform == 'linux' and not util.check_oiejq():
+				print(util.warning('`oiejq` is not installed. It will be installed now.'))
+
+				try:
+					if util.install_oiejq():
+						print(util.info('`oiejq` was successfully installed.'))
+					else:
+						print(util.error('`oiejq` could not be installed. You can try installing it manually.'))
```
tonowak, line 42: This error message won't help the user, unless we tell him how to manually install it. We can at least tell from where to get oiejq and that it's possible to pass oiejq path through a CLI argument.

MasloMaslane:
The problem with testing this functionality is that our runner doesn't have access to https://oij.edu.pl/, and that's where `oiejq.tar.gz` is. I guess we could use normal GitHub runners for testing this.

tonowak:
Sure, normal GitHub runners will be good enough.

src/sinol_make/__init__.py
MasloMaslane: 1a2ba408e2689a8680f91b87a38cd36c7b7647d4

DietPawel:
I am not sure You are all aware, but the oiejq is distributed as a [debian package](https://fox.dasie.mimuw.edu.pl/zaw-repo/pool/contrib/o/oiejq/oiejq_0.1.2_all.deb), which can be installed in ubuntu as well with just dpkg (or even apt install after adding our repo).

PR #24, opened 2023-06-15
Fix execution status order

MasloMaslane:
Before the status for group was the first not OK status received from executing solutions on individual tests. Now the group will have status based on this priority: TLE, MLE, RTE, WA, OK.


MasloMaslane:
Wrong destination branch

PR #25, opened 2023-06-15
Fix execution status order

MasloMaslane:
Before the status for group was the first not OK status received from executing solutions on individual tests. Now the group will have status based on this priority: TLE, MLE, RTE, WA, OK.

### DIFF
--- a/tests/packages/vso/prog/vso2.cpp
+++ b/tests/packages/vso/prog/vso2.cpp
@@ -1,0 +1,26 @@
+// First test RE,
+// second WA,
+// other OK.
+// Group should have status RE.
+
+#include <bits/stdc++.h>
+
+using namespace std;
+
+int main() {
+    int a, b;
+    cin >> a >> b;
+
+    if (a == 1 && b == 1) {
+        int c = 0;
+		cout << a + b / c;
+		return 1;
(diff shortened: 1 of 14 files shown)


MasloMaslane:
Maybe I should add tests for this?

MasloMaslane:
I added tests for this.

tests/packages/vso/prog/vso2.cpp
```diff
@@ -0,0 +1,26 @@
+// First test RE,
+// second WA,
+// other OK.
+// Group should have status RE.
```
tonowak, line 4: Could you also add tests where the final group status is different from the first test status? E.g. so that the tests are OK, OK, WA, OK, RE, OK, WA, and the final status should be RE.

tonowak (commented):
The logic looks correct. I left a comment to improve the tests a bit.

tests/packages/vso/prog/vso2.cpp
MasloMaslane: I added more tests 1f00fa795ab50f147d708732511a445b449f323c

tonowak (approved):
Looks good!

PR #31, opened 2023-06-21
Add workflow for macOS and minor improvements in Ubuntu workflow

MasloMaslane:


MasloMaslane:
The changes in Ubuntu workflow should make the test run faster, but I couldn't test it yet since our runner is currently down

tonowak (approved):
Looks good.

Issue #26, opened 2023-06-20
Alternative time/memory units in config


tonowak:
Currently it's possible to specify the time limit and memory limit in ms and kb, but one could make a mistake in the three zeros. It's also a bit not readable -- it takes a bit of time to understand that `16000ms` is `16s` and not e.g. `1.6s`.

What I propose:
- leave the possibility of specifying in ms and kB,
- add a new way to specify time and memory: `memory_limit_mb` and `time_limit_s`, which takes consecutively an integer and a float,
- implement the same in sio2,
- assert that when someone specifies e.g. both `memory_limit` and `memory_limit_mb`, then their values are equal (e.g. `1000` and `1`).

tonowak:
That would also solve the issue that some people specify the memory limit in powers of 1000, and some in powers of 1024.

MasloMaslane:
I think that's a good idea, but first I'd like to split sinol package management away from sio2 and sioworkers code. I'll try doing that in the coming months

PR #32, opened 2023-06-21
`run` output improvements

MasloMaslane:
- [ ] https://github.com/sio2project/sinol-make/issues/26
- [x] replace `Time limit (in ms): 1000` with `Time limit: 1000 ms`, `Memory limit (in kb): 16000` with `Memory limit: 16000 kB` (note also that probably it should be `kB` instead of `kb`,
- [x] replace `Task Simple test package (abc)` with `Task: Simple test package (tag: abc)`,
- [x] list (e.g. in a single line) the tests for which the solutions will be run,
- [x] after printing the list, remind the user about how to regenerate tests and that `run` runs on the already generated tests,
- [x] give a warning when `run` runs only on sample test,
- [x] give an error when `run` runs on 0 tests,

### DIFF
--- a/src/sinol_make/commands/run/__init__.py
+++ b/src/sinol_make/commands/run/__init__.py
@@ -94,7 +94,7 @@ class Command(BaseCommand):
 
 
 	def extract_test_no(self, test_path):
-		return os.path.split(os.path.splitext(test_path)[0])[1][3:]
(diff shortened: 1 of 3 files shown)


tonowak:
It's not yet ready for review, correct?

tonowak (approved):

tonowak:
I've reviewed the current code. We can merge it and implement the last feature in another PR.

PR #34, opened 2023-06-27
Add `inwer` command

MasloMaslane:


src/sinol_make/helpers/package_util.py
```diff
@@ -0,0 +1,33 @@
+import os
+
+
+def get_task_id() -> str:
+    return os.path.split(os.getcwd())[-1]
+
+
+def extract_test_no(test_path):
+    return os.path.split(os.path.splitext(test_path)[0])[1][3:]
```
tonowak, line 9: Some comments with examples would be helpful to improve the maintainability of the code. Currently I have to try to deduce what it does based on the function name and compare it with my deduction of what the code does.

tonowak: Also, is it a `no` (number), or a string?

tests/helpers/__init__.py
tonowak, line 1: I'm not familiar with Python, does this file need to exist?

src/sinol_make/commands/inwer/__init__.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def configure_subparser(self, subparser: argparse.ArgumentParser):
+        parser = subparser.add_parser(
+            self.get_name(),
+            help='Verify if input files are correct',
+            description='Verify if input files are correct'
+        )
+
+        parser.add_argument('inwer', type=str, nargs='?',
+                            help='path to inwer executable')
```
tonowak, line 29: Doesn't this have to be `--inwer`, just like the arguments below? Also, it might be better to name it `inwer_path`, so that it is immediately clear what this argument does.

src/sinol_make/commands/inwer/__init__.py
```diff
@@ -0,0 +1,115 @@
...
+        add_compilation_arguments(parser)
+
+    def compile_inwer(self, args: argparse.Namespace):
+        self.inwer_executable, compile_log = inwer_util.compile_inwer(self.inwer, args)
+        if self.inwer_executable is None:
+            print(util.error('Compilation failed.'))
+            os.system("head -c 500 %s" % compile_log)
+            exit(1)
+        else:
+            print(util.info('Compilation successful.'))
```
tonowak, line 43: This seems like a function that will also be used in `ingen` etc. Maybe move it to a helper module?

tonowak: Is there a Python `head` alternative? Using commands from shells/operating systems can have unexpected output or behavior on some shells/operating systems. To have a guaranteed stability, it's preferable to use language features or modules written by other people that are sure to work on each operating system.

src/sinol_make/commands/inwer/__init__.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def verify_test(self, execution: InwerExecution) -> VerificationResult:
+        """
+        Veryfies test and returns results.
```
tonowak, line 47:
```suggestion
        Verifies test and returns results.
```

src/sinol_make/commands/inwer/__init__.py
```diff
@@ -0,0 +1,115 @@
...
+        output_file = os.path.join(output_dir, 'inwer.out')
+
+        command = f'{execution.inwer_exe_path} {execution.test_name} < {execution.test_path} > {output_file}'
+        exit_code = os.system(command)
+
+        return VerificationResult(
+            execution.test_path,
+            exit_code == 0,
+            open(output_file, 'r').read()
+        )
```
tonowak, line 60: There might be some built-in Python functions to execute a command and capture it's output and return code. Using it would shorten the code and reduce the chances that there is a bug.

src/sinol_make/commands/inwer/__init__.py
```diff
@@ -0,0 +1,115 @@
...
+        for test in sorted_tests:
+            results[test] = TestResult(test)
+            executions.append(InwerExecution(test, results[test].test_name, self.inwer_executable))
+
+        table_data = TableData(results, 0)
+        print('Verifying tests...\n\n')
+        with mp.Pool(self.cpus) as pool:
+            for i, result in enumerate(pool.imap(self.verify_test, executions)):
+                table_data.results[result.test_path].update_result(result.valid, result.output)
+                inwer_util.print_view(table_data)
```
tonowak, line 75: I'm not sure, can it print the outputs in a non-sorted way?

src/sinol_make/commands/inwer/__init__.py
```diff
@@ -0,0 +1,115 @@
...
+
+        return results
+
+    def run(self, args: argparse.Namespace):
+        if not util.check_if_project():
+            print(util.warning('You are not in a project directory (couldn\'t find config.yml in current directory).'))
+            exit(1)
```
tonowak, line 82: This looks like an error, not a warning. Is there a reason to make it a warning?

src/sinol_make/commands/inwer/__init__.py
```diff
@@ -0,0 +1,115 @@
...
+                util.exit_with_error(f'Inwer "{args.inwer}" not found.')
+        relative_path = os.path.relpath(self.inwer, os.getcwd())
+        print(f'Verifying with inwer {util.bold(relative_path)}')
+
+        self.cpus = args.cpus or mp.cpu_count()
+        self.tests = package_util.get_tests(args.tests)
+
+        if len(self.tests) == 0:
+            print(util.warning('No tests found.'))
+            exit(0)
```
tonowak, line 99: This also looks like an error, not a warning.

tonowak (commented):
The functionality looks excellent! :100: 

Just some minor suggestions to improve the readability / reduce the changes of a bug:

tests/helpers/test_package_util.py
```diff
@@ -0,0 +1,23 @@
...
+from ..fixtures import *
+from sinol_make.helpers import package_util
+
+
+def test_get_task_id(create_package):
+    assert package_util.get_task_id() == "abc"
+
+
+def test_extract_test_no():
+    assert package_util.extract_test_no("in/abc1a.in") == "1a"
```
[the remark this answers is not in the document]
MasloMaslane: Yes, otherwise this directory wouldn't be considered a module and files from it couldn't be imported. More info: https://docs.python.org/3/reference/import.html#regular-packages

src/sinol_make/commands/inwer/__init__.py
MasloMaslane: This way it's a positional argument, so the usage would be like this: `sinol-make run inwer prog/inwer2.cpp`. But I agree that the name should be different.

tests/helpers/__init__.py
MasloMaslane: Yes, otherwise this directory wouldn't be considered a module and files from it couldn't be imported. More info: https://docs.python.org/3/reference/import.html#regular-packages

src/sinol_make/helpers/package_util.py
MasloMaslane: It's a string, for "in/abc1a.in" it returns "1a". But I don't know how to make this function's name better.

src/sinol_make/commands/inwer/inwer_util.py
```diff
@@ -0,0 +1,106 @@
...
+
+def get_inwer(task_id: str, path = None) -> str or None:
+    """
+    Returns path to inwer executable for given task or None if no inwer was found.
+    """
+    if path is None:
+        inwers = glob.glob(os.path.join(os.getcwd(), 'prog', f'{task_id}inwer*'))
+        if len(inwers) == 0:
+            return None
+        return inwers[0]
```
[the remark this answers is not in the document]
MasloMaslane: Changed: d84a479b83b0c9ecdd770ea039cb27d5ad1ad235

src/sinol_make/commands/inwer/inwer_util.py
```diff
@@ -0,0 +1,106 @@
...
+    compile_log_path = os.path.join(compile_log_dir, 'inwer.compile_log')
+    compile_log = open(compile_log_path, 'w')
+
+    try:
+        if compile.compile(inwer_path, output, compilers, compile_log):
+            return output, compile_log_path
+        else:
+            return None, compile_log_path
+    except CompilationError as e:
+        return None, compile_log_path
```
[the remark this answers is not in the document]
MasloMaslane: I added function for compiling files: 2514d829489b32d3fb341033f4fd22cbd59bd5d0

src/sinol_make/commands/inwer/structs.py
```diff
@@ -0,0 +1,47 @@
...
+            self.test_group = group_number
+
+        self.verified = False
+        self.valid = False
+        self.output = ""
+
+    def update_result(self, valid, output):
+        self.verified = True
+        self.valid = valid
+        self.output = output
```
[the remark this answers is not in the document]
MasloMaslane: Renamed 0094ccc34a63098a318da0564a4a141f4f359d4e

src/sinol_make/commands/inwer/structs.py
```diff
@@ -0,0 +1,47 @@
...
+
+@dataclass
+class TableData:
+    results: dict[str, TestResult]
+    previous_vertical_height: int
```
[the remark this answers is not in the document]
MasloMaslane: I added some comments adad71a2f424076ea616596750f6f08cd9b097de

src/sinol_make/helpers/parsers.py
```diff
@@ -0,0 +1,14 @@
+import argparse
+
+from sinol_make.helpers import compiler
+
+
+def add_compilation_arguments(parser: argparse.ArgumentParser):
+	parser.add_argument('--c_compiler_path', type=str, default=compiler.get_c_compiler_path(),
+						help='C compiler to use (default for Linux and Windows: gcc, default for Mac: gcc-9 or gcc-10)')
+	parser.add_argument('--cpp_compiler_path', type=str, default=compiler.get_cpp_compiler_path(),
+						help='C++ compiler to use (default for Linux and Windows: g++, default for Mac: g++-9 or g++-10)')
```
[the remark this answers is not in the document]
MasloMaslane: Added 85128c53e2931e78e2c11c2a4fef5de6d659aa3a

tests/helpers/test_package_util.py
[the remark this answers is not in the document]
MasloMaslane: It will return `12ca`. I added some comments for this function 29b1859ab287eddfed44a57070adaa0e0a6cc597 and tests b6608de9a788d91adc33b781fa07ce78a96936e5 that should make this clearer.

src/sinol_make/commands/inwer/__init__.py
MasloMaslane: Changed the name of this argument feacfe5b4ec4e81912c702878194fd353942f462

src/sinol_make/commands/inwer/__init__.py
MasloMaslane: I added a function for printing first 500 lines from compile log e94fdfd5705b3701f57f578106d9c15200359b61.

src/sinol_make/commands/inwer/__init__.py
MasloMaslane: I don't think there is a better way of running shell scripts than `os.system` or `subprocess.Popen` (which in this case would look exactly the same).

MasloMaslane: I changed the `inwer` executable name, so now the `inwer` command can be run in parallel: 7f667d24dce48ba6cde38b78869b46f65f3512df.

src/sinol_make/commands/inwer/__init__.py
MasloMaslane: Fixed a5fc5fc623349498dc4ca673c18408e4ae73d251.

src/sinol_make/commands/inwer/__init__.py
MasloMaslane: Fixed a5fc5fc623349498dc4ca673c18408e4ae73d251

src/sinol_make/commands/inwer/__init__.py
MasloMaslane: I don't understand what you mean. Every time the table is printed, the cursor moves to the top of the table and prints over the old table.

src/sinol_make/helpers/package_util.py
tonowak: `extract_test_name_without_tag`, or `extract_test_id`? We can also add comments with examples so that it's clear.

src/sinol_make/commands/inwer/__init__.py
tonowak: Nice!

src/sinol_make/commands/inwer/__init__.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def configure_subparser(self, subparser: argparse.ArgumentParser):
+        parser = subparser.add_parser(
+            self.get_name(),
+            help='Verify if input files are correct',
+            description='Verify if input files are correct'
+        )
+
+        parser.add_argument('inwer_path', type=str, nargs='?',
+                            help='path to inwer file')
```
tonowak, line 29: Let's specify it more that it's a file named e.g. `abcinwer.cpp` so that it's clear what is an "inwer file". Because honestly, it's not clear what it is. Maybe we could also name it "inwer source code" instead of "inwer file".

src/sinol_make/commands/inwer/__init__.py
tonowak: > This way it's a positional argument

I see. Though when it's the only positional argument (especially that there are none such arguments in `run`), it seems inconsistent. Now the user needs to know that there are two kinds of arguments. With the big difference between the count of positional and non-positional, it might be better to have only non-positional arguments, thanks to which the user won't feel that "something is weirdly different", just like I did.

src/sinol_make/commands/inwer/__init__.py
```diff
@@ -0,0 +1,113 @@
...
+        parser.add_argument('--tests', type=str, nargs='+',
+                            help='test to verify, for example in/abc{0,1}*')
```
tonowak, line 31: Can it be a regex? Seems important enough to mention it.

src/sinol_make/commands/inwer/__init__.py
```diff
@@ -0,0 +1,113 @@
...
+        parser.add_argument('--cpus', type=int,
+                            help=f'number of cpus to use, you have {mp.cpu_count()} available')
```
tonowak, line 33: What's the default value?

src/sinol_make/commands/inwer/__init__.py
```diff
@@ -0,0 +1,113 @@
...
+        if self.inwer_executable is None:
+            print(util.error('Compilation failed.'))
+            compile.print_compile_log(compile_log_path)
+            exit(1)
+        else:
+            print(util.info('Compilation successful.'))
+
+    def verify_test(self, execution: InwerExecution) -> VerificationResult:
+        """
+        Verifies test and returns results.
```
tonowak, line 47:
```suggestion
        Verifies a test and returns the result of inwer on this test.
```

src/sinol_make/commands/inwer/inwer_util.py
```diff
@@ -0,0 +1,91 @@
...
+import argparse
+
+from sinol_make import util
+from sinol_make.commands.inwer import TestResult, TableData
+from sinol_make.helpers import compile, package_util
+from sinol_make.helpers import compiler
+from sinol_make.interfaces.Errors import CompilationError
+
+
+def get_inwer(task_id: str, path = None) -> str or None:
```
tonowak, line 14:
```suggestion
def get_inwer_path(task_id: str, path = None) -> str or None:
```
Some additional changes might be needed for this one.

src/sinol_make/helpers/package_util.py
```diff
@@ -0,0 +1,51 @@
...
+    if arg_tests is None:
+        all_tests = ["in/%s" % test for test in os.listdir("in/")
+                     if test[-3:] == ".in"]
+        return sorted(all_tests, key=get_test_key)
+    else:
+        return sorted(list(set(arg_tests)), key=get_test_key)
+
+
+def get_file_name(file_path):
+    return os.path.split(file_path)[1]
```
tonowak, line 36: This one is quite situational, right? It might be beneficial to leave a comment (or change the name) about when to use this function.

tonowak (commented):
Hopefully last comments.

src/sinol_make/commands/inwer/__init__.py
MasloMaslane: The parameters are parsed by shell and not the program, so it can be anything that the shell allows

src/sinol_make/helpers/package_util.py
MasloMaslane: No, the function `os.path.split` always splits on the last `/`, so for example for `/some/path/to/file.txt` it will return `['/some/path/to', 'file.txt']`

src/sinol_make/commands/inwer/__init__.py
```diff
@@ -0,0 +1,113 @@
...
+    """
+
+    def get_name(self):
+        return "inwer"
+
+    def configure_subparser(self, subparser: argparse.ArgumentParser):
+        parser = subparser.add_parser(
+            self.get_name(),
+            help='Verify if input files are correct',
+            description='Verify if input files are correct'
```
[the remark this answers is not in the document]
MasloMaslane: I added a better description 9c22ee1cb16d05dbc6dd62757dc415656892606c

src/sinol_make/commands/inwer/__init__.py
MasloMaslane: 9c22ee1cb16d05dbc6dd62757dc415656892606c

src/sinol_make/commands/inwer/__init__.py
MasloMaslane: The default is `mp.cpu_count()`: 9c22ee1cb16d05dbc6dd62757dc415656892606c

src/sinol_make/helpers/package_util.py
MasloMaslane: I changed the name 9c22ee1cb16d05dbc6dd62757dc415656892606c

tonowak (approved):
Let's merge it! :rocket:

tonowak (changes requested):
A comment in the tests have made me realize something -- is the current inwer handling backwards-compatible? New inwers might require passing the filename as a commandline argument, but sio2 doesn't pass them. The inwer will fail in sio2 and the problem package won't upload successfully. This change also requires additional changes in sio2, right? If yes, it might be better to split this PR into two, merge the changes that don't require changing sio2 and wait with merging the breaking changes until changes in sio2 will be implemented.

MasloMaslane:
I created draft PR for that #47, so I think we can merge this after the tests pass

MasloMaslane:
I added support for inwers using asserts and tests for that in #36

src/sinol_make/commands/inwer/__init__.py
```diff
@@ -49,15 +50,17 @@ def verify_test(self, execution: InwerExecution) -> VerificationResult:
...
         os.makedirs(output_dir, exist_ok=True)
-        output_file = os.path.join(output_dir, package_util.get_file_name_without_extension(execution.inwer_exe_path) + ".out")
 
-        command = f'{execution.inwer_exe_path} < {execution.test_path} > {output_file}'
-        exit_code = os.system(command)
+        command = f'{execution.inwer_exe_path} < {execution.test_path}'
+        process = subprocess.Popen(command, shell=True, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
+        process.wait()
+        exit_code = process.returncode
+        out, _ = process.communicate()
```
tonowak, line 58: Does `out` also capture `stderr`? The `assert` output should be included in the output, so that the user would know which assert etc. has been triggered.

MasloMaslane: Yes, there is `stderr=subprocess.STDOUT`

tonowak (approved):

PR #36, opened 2023-06-28
Add tests for `inwer` command

MasloMaslane:


tonowak (commented):
Looks good.

Is the functionality for automatically passing the testname to the inwer already implemented? I haven't checked whether it's implemented, but definitely I haven't found tests for this functionality here.

MasloMaslane:
It is, I just didn't have any ideas how to test it

tonowak:
What about an inwer that checks whether it's filename is the same as the test content, e.g. tests with a string "abc1a.in"?

tonowak (approved):

PR #38, opened 2023-06-30
Compilation improvements

MasloMaslane:


src/sinol_make/commands/run/__init__.py
```diff
@@ -191,8 +193,14 @@ def compile(self, solution):
             self.COMPILATION_DIR, "%s.compile_log" % self.extract_file_name(solution))
         source_file = os.path.join(os.getcwd(), "prog", self.get_solution_from_exe(solution))
         output = os.path.join(self.EXECUTABLES_DIR, self.get_executable(solution))
+
+        weak_compilation_flags = False
+        if hasattr(self, "args") and self.args.weak_compilation_flags:
+            weak_compilation_flags = True
```
tonowak, line 199:
```suggestion
        weak_compilation_flags = hasattr(self, "args") and self.args.weak_compilation_flags
```

tonowak: Also, can `self` not have `args`?

tonowak (commented):
Looks good, my only suggestions:

src/sinol_make/commands/run/__init__.py
MasloMaslane: Changed 7d050b88495eb66ddda18314deefaeca167f4b01

tonowak (commented):
Are there any tests? E.g. ones that check whether the new sinol-make flag indeed compiles some codes that fail to compile on full flags.

MasloMaslane:
No, I'll add one

tonowak (approved):
Nice, let's release a new sinol-make version!

PR #40, opened 2023-07-03
Add support for package with checker in `run`

MasloMaslane:


src/sinol_make/commands/run/__init__.py
```diff
@@ -203,9 +203,59 @@ def compile(self, solution):
             return False
 
 
-    def execute_oiejq(self, command, result_file, output_file, answer_file, time_limit, memory_limit):
+    def check_output_diff(self, output_file, answer_file):
+        """
+        Checks if the output file and the answer file are the same.
+        Returns True if they are the same, False otherwise.
```
tonowak, line 209:
```suggestion
        Checks whether the output file and the answer file are the same.
```
Instead of the comment about returning, it's simpler to add the information that this function is returning a boolean to its signature.

src/sinol_make/commands/run/__init__.py
```diff
@@ -203,9 +203,59 @@ def compile(self, solution):
...
+        """
+        return not os.system("diff -q -Z %s %s >/dev/null" % (output_file, answer_file))
```
tonowak, line 211: If we can avoid using bash commands, we should not use them and instead use the Python alternatives. We don't know on what kind of shells the user is going to run sinol-make. Some commands differ between shells or operating systems. Unless we have a guarantee that for each operating system the `diff` command exists, accepts those arguments, has the same definition of those arguments and returns the same, it's not safe to use the `diff` command.

src/sinol_make/commands/run/__init__.py
```diff
@@ -203,9 +203,59 @@ def compile(self, solution):
...
+
+            return False, 0
+        else:
+            raise CheckerOutputException("Checker output is invalid.")
+
+
+    def check_output(self, name, input_file, output_file, answer_file):
+        """
+        Checks if the output file is correct.
+        Returns a tuple (correct, number of points).
```
tonowak, line 246:
```suggestion
        Returns a tuple (is correct, number of points).
```

src/sinol_make/commands/run/__init__.py
```diff
@@ -495,11 +561,30 @@ def print_expected_scores(self, expected_scores):
...
+                new_expected_scores[solution] = {
+                    "expected": results[solution],
+                    "points": self.calculate_points(results[solution])
+                }
+        else:
+            for solution in results.keys():
+                new_expected_scores[solution] = {
+                    "expected": remove_points_from_results(results[solution]),
+                    "points": self.calculate_points_checker(results[solution])
+                }
```
tonowak, line 587: I'm not sure, what the new code does and why is it needed to split the code into two cases?

src/sinol_make/commands/run/__init__.py
```diff
@@ -594,8 +682,12 @@ def warn_if_not_empty(set, message):
         warn_if_not_empty(diff.removed_groups, "Groups were removed")
 
         for change in diff.changes:
-            print(util.warning("Solution %s passed group %d with status %s while it should pass with status %s." %
-                               (change.solution, change.group, change.result, change.old_result)))
+            if isinstance(change.result, str):
+                print(util.warning("Solution %s passed group %d with status %s while it should pass with status %s." %
+                                   (change.solution, change.group, change.result, change.old_result)))
+            elif isinstance(change.result, int):
+                print(util.warning("Solution %s passed group %d with %d points while it should pass with %d points") %
```
tonowak, line 689:
```suggestion
                print(util.warning("Solution %s passed group %d with %d points while it should pass with %d points.") %
```

src/sinol_make/commands/run/__init__.py
```diff
@@ -594,8 +682,12 @@ def warn_if_not_empty(set, message):
...
+                      (change.solution, change.group, change.result, change.old_result))
```
tonowak, line 690: The indentation here is different than three lines above.

src/sinol_make/commands/run/__init__.py
```diff
@@ -753,11 +845,47 @@ def run(self, args):
             print(util.warning("WARN: Scores sum up to %d (instead of 100)." % total_score))
         print()
 
+        checker = glob.glob(os.path.join(os.getcwd(), "prog", f'{self.ID}chk.*'))
+        if len(checker) != 0:
+            print(util.info("Checker found. Running correct solution first."))
```
tonowak, line 850: Is there a reason for running first the correct solution only when there is a checker? I can't see the scenario where it's needed.

tonowak (commented):
As there have been some task packages with checkers on ONTAK, I'm excited to merge this PR!

src/sinol_make/commands/run/__init__.py
MasloMaslane: The file with a correct answer is one of the arguments given to checker, so first the program has to generate them

src/sinol_make/commands/run/__init__.py
MasloMaslane: The `sinol_expected_scores` for packages with checkers look like this:
```yaml
  chk2.cpp:
    expected:
      1: {points: 25, status: OK}
      2: {points: 25, status: OK}
    points: 50
```

So now there is information about how much points solution scored on each test. Because of this the dictionaries look slightly different and the score calculation is different

src/sinol_make/commands/run/__init__.py
MasloMaslane: I added a function that should work the same as `diff`: b536c3ec6493d57647886b0b0f99ae513c2786c2

src/sinol_make/commands/run/__init__.py
MasloMaslane: Fixed 004c72c88908b8b72be562269013c9a31d999554

src/sinol_make/commands/run/__init__.py
tonowak: Isn't that the job of `make outgen`?

src/sinol_make/commands/run/__init__.py
tonowak: In the future we'll want to add the cutting of solution's scores when they are near the time limit, just like on OI contests in sio2. Such cases will also have status OK with partial points, but that won't be all cases. As such, it would be reasonable to print `{points: 25, status: OK}` only when the `points` are not equal to 100. This way of generating the output won't have to have different cases for the checkers.

src/sinol_make/commands/run/__init__.py
MasloMaslane: Yes it is, I'll change it

src/sinol_make/commands/run/__init__.py
MasloMaslane: I changed how the points are calculated and saved now

src/sinol_make/commands/run/__init__.py
```diff
@@ -203,9 +204,57 @@ def compile(self, solution):
...
+            if len(checker_output) >= 3:
+                try:
+                    points = int(checker_output[2].strip())
+                except ValueError:
+                    pass
+
+            return True, points
+        elif checker_output[0].strip() == "WRONG":
+
+            return False, 0
```
tonowak, line 238:
```suggestion
            return False, 0
```

src/sinol_make/commands/run/__init__.py
```diff
@@ -562,11 +644,33 @@ def validate_expected_scores(self, results):
...
+                            group=field[2],
+                            old_points=change[0]["points"],
+                            new_points=change[1]["points"]
+                        ))
+                    else:
+                        solution = field[0]
+                        group = field[2]
+                        old_result = change[0]
+                        result = change[1]
+                        changes.append(ResultChange(solution, group, old_result, result))
```
tonowak, line 673: Inconsistent styles in the branches. Also, the `solution` and `group` is always the same, so maybe it's possible to shorten the current code?

src/sinol_make/commands/run/__init__.py
```diff
@@ -594,8 +698,16 @@ def warn_if_not_empty(set, message):
...
+            if isinstance(change, ResultChange):
+                if isinstance(change.result, str):
+                    print(util.warning("Solution %s passed group %d with status %s while it should pass with status %s." %
+                                       (change.solution, change.group, change.result, change.old_result)))
+                elif isinstance(change.result, int):
+                    print(util.warning("Solution %s passed group %d with %d points while it should pass with %s points." %
+                                       (change.solution, change.group, change.result, change.old_result)))
+            elif isinstance(change, PointsChange):
+                print(util.warning("Solution %s passed group %d with %d points while it should pass with %d points." %
+                                   (change.solution, change.group, change.new_points, change.old_points)))
```
tonowak, line 710: Those two strings are the same, should they be? If yes, maybe we can shorten the code?

src/sinol_make/commands/run/__init__.py
```diff
@@ -753,11 +865,34 @@ def run(self, args):
...
+            if not checker_compilation[0]:
+                util.exit_with_error('Checker compilation failed.')
+
+            ins = glob.glob(os.path.join(os.getcwd(), "in", f'{self.ID}*.in'))
+            outs = glob.glob(os.path.join(os.getcwd(), "out", f'{self.ID}*.out'))
+
+            ins = set([self.extract_test_no(test) for test in ins])
+            outs = set([self.extract_test_no(test) for test in outs])
+            if ins != outs:
+                util.exit_with_error('In and out files do not match.')
```
tonowak, line 892:
```suggestion
            if ins != outs:
                util.exit_with_error('The list of names of input files is different than of output files.')
```

src/sinol_make/commands/run/__init__.py
MasloMaslane: They shouldn't be

src/sinol_make/commands/run/__init__.py
MasloMaslane: Refactored c3c7c62958dfcbbee405ba26d7d4269907a695a4

src/sinol_make/commands/run/__init__.py
MasloMaslane: Sorry, they should, i shortened the code: c01a6932b6816a2df51ad94039e2eab099d99597

tonowak (approved):
:rocket:

PR #48, opened 2023-07-04
Split points evenly when `scores` not defined in `config.yml`

### LINKED ISSUE #44: Split points evenly across groups when they are not defined in config
The points should be split evenly across groups when `scores` is not defined in `config.yml`

MasloMaslane:
Closes #44

### DIFF
--- a/src/sinol_make/commands/run/__init__.py
+++ b/src/sinol_make/commands/run/__init__.py
@@ -474,10 +474,10 @@ class Command(BaseCommand):
     def calculate_points(self, results):
         points = 0
         for group, result in results.items():
-            if group != 0 and group not in self.config["scores"]:
+            if group != 0 and group not in self.scores:
                 util.exit_with_error(f'Group {group} doesn\'t have points specified in config file.')
             if result == "OK" and group != 0:
-                points += self.config["scores"][group]
+                points += self.scores[group]
         return points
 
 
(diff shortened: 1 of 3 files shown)


src/sinol_make/commands/run/__init__.py
```diff
@@ -706,6 +706,42 @@ def validate_arguments(self, args):
...
+
+        self.possible_score = self.get_possible_score(self.groups)
+
+        print("Scores:")
+        total_score = 0
+        for group in self.scores:
+            print("%2d: %3d" % (group, self.scores[group]))
+            total_score += self.scores[group]
+        if total_score != 100:
+            print(util.warning("WARN: Scores sum up to %d (instead of 100)." % total_score))
```
tonowak, line 742:
```suggestion
            print(util.warning("WARN: Scores sum up to %d instead of 100." % total_score))
```

src/sinol_make/commands/run/__init__.py
```diff
@@ -706,6 +706,42 @@ def validate_arguments(self, args):
...
+
+            if points_per_group * num_groups != 100:
+                self.scores[self.groups[-1]] += 100 - points_per_group * num_groups
+        else:
+            for group in self.config["scores"]:
+                self.scores[group] = self.config["scores"][group]
+
+        self.possible_score = self.get_possible_score(self.groups)
+
+        print("Scores:")
```
tonowak, line 736: Does the scores get printed each time `sinol-make run` is run? I would suggest to avoid printing too much data, especially those that are easily accessible (e.g. by reading `config.yml`) and not produced by `sinol-make`.

MasloMaslane: So maybe the program should print it only when `scores` is not defined in config?

tonowak: A suitable warning could be enough (one that e.g. explains what has changed and where to look for the changes). But if you think it would be better to print the whole `scores`, go for it.

src/sinol_make/commands/run/__init__.py
```diff
@@ -706,6 +706,42 @@ def validate_arguments(self, args):
...
+            if points_per_group * num_groups != 100:
+                self.scores[self.groups[-1]] += 100 - points_per_group * num_groups
+
+            print("Points will be assigned as follows:")
+            total_score = 0
+            for group in self.scores:
+                print("%2d: %3d" % (group, self.scores[group]))
+                total_score += self.scores[group]
+            if total_score != 100:
+                print(util.warning("WARN: Scores sum up to %d instead of 100." % total_score))
```
tonowak, line 737: This warning isn't reachable, right? It should be printed when the user's own configuration is wrong, not the one provided by us.

MasloMaslane: Yes, I changed that

tonowak (approved):

PR #50, opened 2023-07-04
Add flag to exit on failed compilation

### LINKED ISSUE #43: Run even if compilation of some files failed
When some files don't compile, sinol-make shouldn't exit but try running on compiled files

MasloMaslane:
Closes #43

### DIFF
--- a/src/sinol_make/commands/run/__init__.py
+++ b/src/sinol_make/commands/run/__init__.py
@@ -61,6 +61,8 @@ class Command(BaseCommand):
                             help='use weaker compilation flags')
         parser.add_argument('--apply_suggestions', dest='apply_suggestions', action='store_true',
                             help='apply suggestions from expected scores report')
+        parser.add_argument('--ignore_compilation_errors', dest='ignore_compilation_errors', action='store_true',
+                            help='run compiled solutions, even if some of them did not compile')
 
 
     def color_memory(self, memory, limit):
@@ -183,7 +185,7 @@ class Command(BaseCommand):
         print("Compiling %d solutions..." % len(solutions))
         with mp.Pool(self.cpus) as pool:
             compilation_results = pool.map(self.compile, solutions)
-        if not all(compilation_results):
+        if not self.args.ignore_compilation_errors and not all(compilation_results):
             util.exit_with_error("\nCompilation failed.")
         return compilation_results
 
@@ -483,6 +485,14 @@ class Command(BaseCommand):
 
     def compile_and_run(self, solutions):
         compilation_results = self.compile_solutions(solutions)
+        offset = 0
+        for i in range(len(solutions)):
+            if not compilation_results[i]:
+                self.failed_compilations.append(solutions[i - offset])
+                solutions.pop(i - offset)
+                offset += 1
+        compilation_results = [result for result in compilation_results if result]
+
         os.makedirs(self.EXECUTIONS_DIR, exist_ok=True)
         executables = [os.path.join(self.EXECUTABLES_DIR, self.get_executable(solution))
                        for solution in solutions]
@@ -509,6 +519,11 @@ class Command(BaseCommand):
         used_solutions = results.keys()
         if self.args.solutions == None and config_expected_scores: # If no solutions were specified, use all solutions from config
             used_solutions = config_expected_scores.keys()
+        used_solutions = list(used_solutions)
+
+        for solution in self.failed_compilations:
+            if solution in used_solutions:
+                used_solutions.remove(solution)
 
         used_groups = set()
         if self.args.tests == None and config_expected_scores: # If no groups were specified, use all groups from config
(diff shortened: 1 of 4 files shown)


tonowak:
To be nicer for the users, ignoring the compilation errors can be the default behavior, but then for the CI and scripts, `sinol-make run` should return with a non-zero code in case of compilation errors. This method avoids adding a new flag and doesn't annoy the user in any way.

MasloMaslane:
I changed the behaviour, now by default sinol-make runs with compiled solutions and there is a flag for exiting on failed compilation

MasloMaslane:
I couldn't come up with any scenario, so I'll change it now

MasloMaslane:
I changed it

src/sinol_make/commands/run/__init__.py
```diff
@@ -183,8 +183,8 @@ def compile_solutions(self, solutions):
         print("Compiling %d solutions..." % len(solutions))
         with mp.Pool(self.cpus) as pool:
             compilation_results = pool.map(self.compile, solutions)
-        if not all(compilation_results):
-            util.exit_with_error("\nCompilation failed.")
+        if len(list(filter(lambda x: x, compilation_results))) == 0:
+            util.exit_with_error("\nNo solutions were compiled.")
```
tonowak, line 187: I'm not sure whether I understand this code. Sinol-make should exit with an error when at least one solution failed to compiled, not when all solutions failed to compile. That should be done at the end, after printing the table etc.

tonowak: The reasoning is that we want to me kind as much as possible and print the table even when something is wrong with the solutions, but nevertheless we should return with a non-zero exit code so that CI etc. doesn't pass.

MasloMaslane: When some solutions fail to compile the code will exit with error at the end ([here](https://github.com/sio2project/sinol-make/blob/9b9ad252afc407846bcf4a06d294715461e329c6/src/sinol_make/commands/run/__init__.py#L722)), but this two lines exit when every solution had compilation errors

MasloMaslane: So when no solution compiles successfully, the executions won't start

tonowak: Is there a reason for handling this edge-case? E.g. does the output look bad for some reason when no solution compiled? As long as it doesn't help the user, we should try to make the code shorted so it's easier to maintain.

MasloMaslane: Maybe this line would be more readable: `if not any(compilation_results):`

MasloMaslane: The output looks like this without the if:
```
Compilation of file wcf.cpp was unsuccessful.
/mnt/c/Users/maslo/oi/sinol-make/tests/packages/wcf/prog/wcf.cpp: In function ‘int main()’:
/mnt/c/Users/maslo/oi/sinol-make/tests/packages/wcf/prog/wcf.cpp:7:9: error: unused variable ‘c’ [-Werror=unused-variable]
    7 |     int c; // Unused variable
      |         ^
cc1plus: all warnings being treated as errors

Performing 0 executions...
Expected scores from config:
sinol_expected_scores: {}


Expected scores based on results:
sinol_expected_scores: {}

Expected scores are correct!
Compilation failed for 1 solution(s).
```

tonowak: Doesn't look necessarily bad, it can stay like this. That would shorten our code and reduce chances of a bug.

If we want to make some if's so that the output is nicer, I would say that we can shorten the "expected scores output". There is no reason to print both tables when they are equal. But that is better left for another PR.

src/sinol_make/commands/run/__init__.py
```diff
@@ -483,6 +481,14 @@ def calculate_points(self, results):
 
     def compile_and_run(self, solutions):
         compilation_results = self.compile_solutions(solutions)
+        offset = 0
+        for i in range(len(solutions)):
+            if not compilation_results[i]:
+                self.failed_compilations.append(solutions[i - offset])
+                solutions.pop(i - offset)
+                offset += 1
```
tonowak, line 489: Just to be sure: this code works when some elements get popped? It's not clear that this is correct. A more readable solution would be to create a second `solutions` list, or something other.

src/sinol_make/commands/run/__init__.py
```diff
@@ -706,6 +717,10 @@ def validate_arguments(self, args):
 
         return compilers, timetool_path
 
+    def exit(self):
+        if len(self.failed_compilations) > 0:
+            util.exit_with_error('Compilation failed for %d solution(s).' % len(self.failed_compilations))
```
tonowak, line 722:
```suggestion
            util.exit_with_error('Compilation failed for {cnt} solution{letter}.'.format(cnt=len(self.failed_compilations), letter='' if self.failed_compilations.empty() else 's'))
```
(this suggestion might have a syntax error, this was hand-written)

tonowak (approved):
Thanks for making this PR! The error on at least one compilation error was annoying for the jury of ONTAK, hopefully now they'll have a better user experience :)

PR #52, opened 2023-07-05
Move to subprocess

MasloMaslane:
`subprocess.Popen` is more flexible than `os.system` and it will make easier to avoid problems such as not escaping spaces.

### DIFF
--- a/src/sinol_make/commands/inwer/__init__.py
+++ b/src/sinol_make/commands/inwer/__init__.py
@@ -53,9 +53,11 @@ class Command(BaseCommand):
         output_dir = os.path.join(os.getcwd(), 'cache', 'executions', execution.test_name)
         os.makedirs(output_dir, exist_ok=True)
 
-        command = f'{execution.inwer_exe_path} < {execution.test_path}'
-        process = subprocess.Popen(command, shell=True, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
+        test = open(execution.test_path, 'r')
+        command = [execution.inwer_exe_path]
+        process = subprocess.Popen(command, stdin=test, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
         process.wait()
+        test.close()
         exit_code = process.returncode
         out, _ = process.communicate()
 
--- a/src/sinol_make/commands/inwer/__init__.py
+++ b/src/sinol_make/commands/inwer/__init__.py
@@ -53,11 +53,10 @@ class Command(BaseCommand):
         output_dir = os.path.join(os.getcwd(), 'cache', 'executions', execution.test_name)
         os.makedirs(output_dir, exist_ok=True)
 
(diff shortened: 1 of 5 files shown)


src/sinol_make/commands/inwer/__init__.py
```diff
@@ -53,9 +53,11 @@ def verify_test(execution: InwerExecution) -> VerificationResult:
         output_dir = os.path.join(os.getcwd(), 'cache', 'executions', execution.test_name)
         os.makedirs(output_dir, exist_ok=True)
 
-        command = f'{execution.inwer_exe_path} < {execution.test_path}'
-        process = subprocess.Popen(command, shell=True, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
+        test = open(execution.test_path, 'r')
+        command = [execution.inwer_exe_path]
+        process = subprocess.Popen(command, stdin=test, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
         process.wait()
+        test.close()
```
tonowak, line 60: Because of readability and error-proneness, it's preferred to open a file with the `with` statement: https://docs.python.org/3/reference/compound_stmts.html#the-with-statement

tonowak: This comment can be applied to other parts of code in this PR.

src/sinol_make/commands/run/__init__.py
```diff
@@ -239,14 +236,11 @@ def execute_oiejq(self, command, input_file_path, output_file_path, answer_file_
...
     def execute_time(self, command, result_file_path, input_file_path, output_file_path, answer_file_path,
                      time_limit, memory_limit):
-        input_file = open(input_file_path, "r")
-        output_file = open(output_file_path, "w")
-        process = subprocess.Popen(command, stdin=input_file, stdout=output_file, stderr=subprocess.DEVNULL)
-        process.wait()
+        with open(output_file_path, "w") as output_file, open(input_file_path, "r") as input_file:
+            process = subprocess.Popen(command, stdin=input_file, stdout=output_file, stderr=subprocess.DEVNULL)
+            process.wait()
+            output_file.flush()
```
tonowak, line 242: Do we have to save the output to a file? I guess that Python handles capturing the output to a variable, without a new file. Using files with absolute paths makes it easier to make a bug by e.g. running two sinol-makes at the same time. In short, when possible, it's better to redirect directly to a variable or a path with a random name, rather than to a file. Especially that the file is not getting deleted and it uses memory in the filesystem.

tonowak: Also, the `flush` here might not be needed, right? I'm not sure, but the file might be flushed on closing.

MasloMaslane: I changed so that `with` statements are now used

tonowak (approved):

PR #54, opened 2023-07-05
Check for sinol-make updates when running

MasloMaslane:


tonowak:
How does it work when there is no internet connection or it is very slow? In such cases, we shouldn't annoy the user in any way.

MasloMaslane:
When there’s no internet the code will work well, but I haven’t thought about slow connections

MasloMaslane:
I added a one second timeout for the request

tonowak:
Can this be reduced to e.g. 0.3s? We should be as non-invasive as possible. In fact, if it's possible to cache when we have last checked for updates, it would be even better, instead of executing it on each `run` command.

src/sinol_make/util.py
```diff
@@ -154,6 +157,69 @@ def save_config(config):
...
+    if not data_dir.is_dir():
+        os.mkdir(data_dir)
+
+    thread = threading.Thread(target=check_version)
+    thread.start()
+    version_file = data_dir.joinpath("version")
+    if version_file.is_file():
+        version = version_file.read_text()
+        try:
+            if compare_versions(current_version, version) == -1:
```
tonowak, line 176: What if we just updated the version? This file probably will still exist (the `check_version` won't be fast enough to overwrite here) and the warning will be printed, even though we have just updated the version.

tonowak (commented):
Can I ask to add a few more comments? It's not immediately obvious what the new code does, what is being done concurrently, why do we have a new directory, what are possible orders of executions.

tonowak:
Can we add tests? I want to be sure that the warning isn't printed when it shouldn't be printed and the code doesn't crash. So we could e.g. test the code when there is no internet connection, or with different release versions (that would require somehow changing the link or the behavior, so that testing it will be possible).

src/sinol_make/util.py
MasloMaslane: Then the current version will be the same (or higher) than the one in the file, so I don't think there will be a problem

tonowak: I see, you are right. Nevertheless, adding comments explaining why this method is correct would improve readability and maintainability.

src/sinol_make/__init__.py
```diff
@@ -32,6 +32,11 @@ def main():
 
     for command in commands:
         if command.get_name() == args.command:
+            new_version = util.check_for_updates(__version__)
+            if new_version is not None:
+                print(util.warning(f'New version of sinol-make is available (your version: {__version__}, available version: {new_version}).\n'
+                                   f' You can update it by running `pip install sinol-make --upgrade`.'))
```
tonowak, line 38:
```suggestion
                                   f' You can update it by running `pip3 install sinol-make --upgrade`.'))
```

tonowak (approved):

PR #55, opened 2023-07-06
Add checks if output files are missing 

### LINKED ISSUE #37: Check if every input file has an output file
In the `run` command, add a check that each input file has a corresponding output file.

MasloMaslane:
Closes #37


src/sinol_make/commands/run/__init__.py
```diff
@@ -747,6 +748,25 @@ def run(self, args):
             example_tests = [test for test in self.tests if self.get_group(test) == 0]
             if len(example_tests) == len(self.tests):
                 print(util.warning('Running only on example tests.'))
+
+            output_tests = glob.glob(os.path.join(os.getcwd(), "out", "*.out"))
```
tonowak, line 752: This function is now very long, would it possible to split in into smaller functions?

MasloMaslane: Sure

src/sinol_make/commands/run/__init__.py
```diff
@@ -700,6 +705,39 @@ def set_scores(self):
...
+        """
+        output_tests = glob.glob(os.path.join(os.getcwd(), "out", "*.out"))
+        output_tests_ids = [self.extract_test_id(test) for test in output_tests]
+        valid_input_files = []
+        for test in self.tests:
+            if self.extract_test_id(test) in output_tests_ids:
+                valid_input_files.append(test)
+        return valid_input_files
+
+    def validate_tests(self):
```
tonowak, line 720:
```suggestion
    def validate_existence_of_outputs(self):
```
The current name is misleading (can be related to inwer) and not descriptive of what the function does. I would say that the function can be split further into smaller functions with better names, which would make the code clearer.

tonowak: In fact, I'm not sure whether the name I've proposed is even accurate.

MasloMaslane: I made some changes, but I didn't come up with a better name for this function

src/sinol_make/commands/run/__init__.py
```diff
@@ -717,24 +717,34 @@ def get_valid_input_files(self):
                 valid_input_files.append(test)
         return valid_input_files
 
+    def validate_existance_of_outputs(self):
```
tonowak, line 720: https://en.wiktionary.org/wiki/existance

src/sinol_make/commands/run/__init__.py
```diff
@@ -717,24 +717,34 @@ def get_valid_input_files(self):
...
+            missing_tests = list(set(self.tests) - set(valid_input_files))
+            missing_tests.sort()
+
+            print(util.warning('Missing output files for tests: ' + ', '.join(
+                [self.extract_file_name(test) for test in missing_tests])))
+            print(util.warning('Running only on tests with output files.'))
+            self.tests = valid_input_files
+            self.groups = self.get_groups(self.tests)
+
     def validate_tests(self):
```
tonowak, line 735:
```suggestion
    def check_are_any_tests_to_run(self):
```
Sometimes it's as simple as shortening the description of the function.

tonowak (approved):
That looks good!

PR #58, opened 2023-07-12
Change flags to be more Linux-style

MasloMaslane:
I also added shorter flags for some commands, let me know what you think


tonowak (approved):
I'm also in favor of `-` instead of `_`.

PR #59, opened 2023-07-12
Add autocompletion

MasloMaslane:
I found a package for autocompletion, but I'd like you to test it. For me the completion works in about a second, so it feels kinda slow. 

After installing sinol-make you can enable it either by adding `eval "$(register-python-argcomplete sinol-make)"` to your `.bashrc` (or `.zshrc`) or using `activate-global-python-argcomplete` and restarting shell.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,13 +46,10 @@ other tools for measuring time (which are non-deterministic and produce reports 
 - Windows (WSL): `apt install time timeout`
 - macOS: `brew install gnu-time coreutils`
 
-If you would like to have autocomplete for `sinol-make` commands,
-add the following line to your `~/.bashrc` (or `~/.zshrc`) file:
+### Autocompletion (optional)
 
-```shell
-eval "$(register-python-argcomplete sinol-make)"
-```
-or run the following command:
+If you would like to have autocompletion for `sinol-make` commands,
+run the following command and refresh the shell (e.g. by opening a new terminal):
 
 ```shell
 activate-global-python-argcomplete
--- a/README.md
+++ b/README.md
@@ -46,6 +46,18 @@ other tools for measuring time (which are non-deterministic and produce reports 
 - Windows (WSL): `apt install time timeout`
 - macOS: `brew install gnu-time coreutils`
 
+If you would like to have autocomplete for `sinol-make` commands,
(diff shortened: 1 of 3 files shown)


tonowak:
It seems to work in half a second for me.

As it's completely backwards compatible, I would say that it's OK to have this feature, even if it is not perfect. When the autocompletion is too slow for someone, one can just not use it.

tonowak (approved):

pyproject.toml
```diff
@@ -20,17 +20,18 @@ classifiers = [
...
 ]
 
 [project.optional-dependencies]
 tests = [
-  "pytest",
-  "pytest-cov",
-  "requests-mock",
+    "pytest",
+    "pytest-cov",
+    "requests-mock",
```
tonowak, line 34: Is this change intentional?

MasloMaslane: Yes, I wanted to have the same indentation everywhere in pyproject.toml

MasloMaslane:
I added installation steps to the Readme

README.md
```diff
@@ -46,6 +46,18 @@ other tools for measuring time (which are non-deterministic and produce reports
 - Windows (WSL): `apt install time timeout`
 - macOS: `brew install gnu-time coreutils`
 
+If you would like to have autocomplete for `sinol-make` commands,
```
tonowak, line 49:
```suggestion
### Autocompletion (optional)

If you would like to have autocompletion for `sinol-make` commands,
```

I'm putting it into another section to reduce the amount of needed installation steps. We don't want anyone to look into the installation steps, think "ugh, it's long, I don't wanna do so many steps" and not try our tool.

README.md
```diff
@@ -46,6 +46,18 @@ other tools for measuring time (which are non-deterministic and produce reports
...
+add the following line to your `~/.bashrc` (or `~/.zshrc`) file:
+
+```shell
+eval "$(register-python-argcomplete sinol-make)"
+```
+or run the following command:
```
tonowak, line 55:
```suggestion
run the following command and refresh the shell (e.g. by opening a new terminal):
```

tonowak: Even though the command requires refreshing the shell, it's easier to do and requires less technical knowledge.

PR #60, opened 2023-07-13
Fix errors with executing solutions

MasloMaslane:
I got rid of bash `timeout` and replaced it with `subprocess`' timeout, because it was causing some problems. I also fixed killing processes when they run longer then the hard time limit


tonowak (approved):
Seems good, thanks for putting it together? Can I know what kind of issues was the old `timeout` producing?

MasloMaslane:
For some reason on slow tests that still should pass with OK, `timeout` would kill the solution because it ran too slow. This problem only occured when using subprocess for some reason. I couldn't find out why it was happening, so I changed from `timeout` to subprocess' timeout

tonowak:
Thanks for the explanation. Feel free to merge this PR :)

PR #61, opened 2023-07-14
Always display both tables in run using new printing function

### LINKED ISSUE #7: Always display both tables in run command
Both tables should be displayed while running `run` command. Currently only small version is displayed.

MasloMaslane:
I wrote a function that prints similarly to `less` command, I like how it turned out. Let me know what you think.

~~When there are too many solutions the output will still break, I'll fix it soon.~~ Fixed

Closes #7


tonowak:
Locally it "flashes" the screen for me. Do you have the same issue?

tonowak:
The "Done" and "Time remaining" doesn't make sense at all -- sometimes there are more done tests than total (e.g. `Done  311/204`) and sometimes the time remaining is negative (e.g. `-167 seconds.`). Feel free to create an issue and fix it in another PR.

tonowak:
I'm wondering whether setting the background color to the terminal's default would help with the blinking of the screen. Or somehow reducing the refresh rates to only when something changes.

tonowak:
I like that it prints the table at the end nevertheless, that's a cool idea  and it definitely should stay!

MasloMaslane:
I fixed the flickering and the time remaining

tonowak:
The flickering is fixed, but now it seems that there are delays while scrolling. I guess you can try to reproduce it by running it on a package with a lot of tests and slow solutions.

tonowak:
I don't know how easy/hard it would be, but an improvement that would raise the readability of the table would be to create horizontal lines after each test group (but not between each test).

tonowak:
Exiting from the view (e.g. by ctrl-C) could also print the table at the end. Currently it's being printed only at the end of a successful run, but it would be nice to see it also after interrupting the run.

tonowak:
By the way, the current table already is incredible. You have no idea how I'm excited I am to have it in sinol-make :)

MasloMaslane:
> The flickering is fixed, but now it seems that there are delays while scrolling. I guess you can try to reproduce it by running it on a package with a lot of tests and slow solutions.

I was trying to find a fix for the delays with scrolling on slow solutions, but haven't succeeded yet. I'm guessing it's because of the slow solutions, but in theory it shouldn't be a problem, since they're running in parallel.

MasloMaslane:
There's also still a problem, when there's a Keyboard Interrupt while slow solutions are running then they aren't killed and still run in the background

tonowak:
> There's also still a problem, when there's a Keyboard Interrupt while slow solutions are running then they aren't killed and still run in the background

Do the solutions get killed after a while? It's OK to have a delay before the processes die, I guess it's harder to kill them immediately.

tonowak:
> I'm guessing it's because of the slow solutions, but in theory it shouldn't be a problem, since they're running in parallel.

You can verify that by running on a single core. The scroll delay is then still present.

MasloMaslane:
I added splitting line between groups, changed so that the table is displayed on keyboard interrupt, fixed killing of slow solutions and the amount of how many programs can be displayed is calculated dynamically (you can resize your terminal and check)

MasloMaslane:
And I haven't fixed the responsiveness on slow solutions

MasloMaslane:
I fixed the error with scrolling, it was just a small bug. I also added a title and footer for the printer, let me know what you think.

tonowak:
That is looking excellent! I have some slight suggestions to make it even better, after which hopefully we can merge it. Each of them is a small suggestion, probably they won't take much time to implement:
- to make it a bit easier for the user to read the text, let's add margins: an additional, empty line (or a line with borders of the table) at the top and at the bottom of the table, and also additional single spaces (or maybe even two spaces) at the left and at the right as the letters are too close to the end of the terminal and it's a bit hard to read such letters,
- the horizontal lines that separate test groups don't show correctly when there is too much solutions to fit in one table,
- on my terminal the table has a grayish background with a white/gray font; if you want this color of the background, the font needs to be white, not grayish, otherwise they blend too much and it takes some effort to read the letters,
- the end of the table is missing a horizontal line, which would tell the user that the table definitely has ended,
- let's also put a horizontal line between the summary and the sample test group, as currently the sample tests seem to be a part of the summary.

I haven't yet looked at the code, but I think I won't have much comments about it.

It's hard to write some tests for this feature, but I think it would be enough to manually check that this feature work on the operating systems that we care about. Are you able to somehow verify it on e.g. on a Windows or on a Mac virtual machine that the table shows correctly and the user can correctly interact with it?

MasloMaslane:
I added the margin, horizontal line at the end and between test group and summary, and the color of the table should be now white text on black background.

I don't see any issues with line separators between groups.

MasloMaslane:
The code works well both on Windows WSL and macOS

tonowak:
> the color of the table should be now white text on black background.

The colors look the same for me :(

MasloMaslane:
What terminal and theme do you use? Now the code specifies to use white and black, before it should have been the default color

tonowak:
Alacritty with the following configutation (that also has the color codes I'm using): https://pastebin.com/WckPtRA7

MasloMaslane:
I’ll try testing it tomorrow

MasloMaslane:
I fixed the problem with background colors

tonowak:
Nice, the colors looks good now!

> I don't see any issues with line separators between groups.

I cannot reproduce it anymore :( hopefully it somehow got fixed.

src/sinol_make/commands/run/__init__.py
```diff
@@ -397,135 +575,51 @@ def run_solutions(self, compiled_commands, names, solutions, report_file):
...
+
+        if has_terminal:
+            run_event = threading.Event()
+            run_event.set()
+            thr = threading.Thread(target=printer.printer_thread, args=(run_event, print_view, program_groups_scores,
+                                                                        all_results, print_data, names, executions,
+                                                                        self.groups, self.scores, self.tests,
+                                                                        self.possible_score, self.time_limit,
+                                                                        self.memory_limit, self.cpus,
+                                                                        self.args.hide_memory))
```
tonowak, line 597: That's not the ideal intendation, I guess.

src/sinol_make/helpers/printer.py
```diff
@@ -0,0 +1,179 @@
+import curses
+
+from datetime import datetime, timedelta
+from curses import wrapper
+
+
+def printer(func, *args, **kwargs):
+    """
+    Prints output of func (called every 0.1 seconds with terminal width and height, then args and kwargs)
```
tonowak, line 9: Is there a reason for choosing specifically 0.1? Maybe we can refresh it faster? Probably that would smoothen out the scroll, though it's not necessary.

src/sinol_make/helpers/printer.py
```diff
@@ -0,0 +1,179 @@
...
+
+                stdscr.refresh()
+            last_output = output
+            last_title = title
+            last_footer = footer
+    except KeyboardInterrupt:
+        return
+
+
+def _print_to_scr(scr: curses.window, output, title, footer):
```
tonowak, line 124: The code quality of this function is, well, not perfect. As long as we won't need to change this function (or even look at it), I'm OK with merging it, but having some comments or explanations wouldn't hurt, just in case. I leave it up to you -- I'm fine with merging it as it is.

tonowak (commented):
Non-blocking comments from code review:

src/sinol_make/helpers/printer.py
MasloMaslane: The `func`'s output will change every 0.1 seconds, but the keyboard input is polled as often as possible, so scrolling should be fast. We could change the 0.1 second interval. We can reduce the time interval, but I don't think it will change much visually, but it could slow down scrolling.

src/sinol_make/commands/run/__init__.py
MasloMaslane: I changed the intendation.

src/sinol_make/helpers/printer.py
MasloMaslane: I added some comments, what do you think?

src/sinol_make/commands/run/__init__.py
tonowak: I haven't checked, but the lines still seem to be over 120 characters long. Do they fit, or are they over 120?

src/sinol_make/helpers/printer.py
tonowak: Yeah, it seems better with those comments, thanks!

tonowak (approved):
This table will drastically improve the satisfaction of the users :)

src/sinol_make/commands/run/__init__.py
MasloMaslane: They're under 120 lines

tonowak: characters* not lines :sweat_smile:

MasloMaslane: Yes, that's what I meant

PR #62, opened 2023-07-17
Use printer in inwer

MasloMaslane:
I also added --weak_compilation_flags to inwer.


src/sinol_make/commands/inwer/__init__.py
```diff
@@ -74,12 +87,39 @@ def verify_and_print_table(self) -> dict[str, TestResult]:
...
             executions.append(InwerExecution(test, results[test].test_name, self.inwer_executable))
 
+        has_terminal = True
+        try:
+            terminal_width = os.get_terminal_size().columns
+            terminal_height = os.get_terminal_size().lines
+        except OSError:
+            has_terminal = False
+            terminal_width = 80
+            terminal_height = 30
```
tonowak, line 97: That looks like code that probably is also in `sinol-make run` (though I haven't checked). Can this be put into a separate function?

tonowak: In short, it would be best to move the terminal's logic as much as possible out of `inwer`.

MasloMaslane: I added a function for this

tonowak: Maybe other code can also be moved out of `inwer` and shared with `run`?

MasloMaslane: I looked a bit at the code but didn't find anything

tonowak (approved):

PR #63, opened 2023-07-23
Input and output file generation command 

MasloMaslane:


src/sinol_make/commands/gen/gen_util.py
```diff
@@ -0,0 +1,124 @@
...
+
+def run_ingen(ingen_exe):
+    """
+    Runs ingen and generates all input files.
+    :param ingen_exe: path to ingen executable
+    :return: True if ingen was successful, False otherwise
+    """
+    shell = False
+    if os.path.splitext(ingen_exe)[1] == '.sh':
+        shell = True
```
tonowak, line 79:
```suggestion
    shell = os.path.splitext(ingen_exe)[1] == '.sh'
```
Also, `is_shell` might be a better name, as currently the name suggest that this variable contains the shell type, or something similar.

src/sinol_make/commands/gen/__init__.py
```diff
@@ -0,0 +1,147 @@
...
+            help='Generate input and output files',
+            description='Generate input files using gingen program '
+                        '(for example prog/abcingen.cpp for abc task). If there '
+                        'is a new test or it differs from the previous one, '
+                        'then the output file will be generated using the '
+                        'correct solution. You can also specify your ingen source '
+                        'file which will be used.'
+        )
+
+        parser.add_argument('ingen_path', type=str, nargs='?',
```
tonowak, line 36: What `nargs='?'` means?

src/sinol_make/commands/gen/__init__.py
```diff
@@ -0,0 +1,147 @@
...
+
+    def compile_ingen(self):
+        self.ingen_exe, compile_log_path = gen_util.compile_ingen(self.ingen, self.args,
+                                                                  self.args.weak_compilation_flags)
+        if self.ingen_exe is None:
+            print(util.error('Failed ingen compilation.'))
+            compile.print_compile_log(compile_log_path)
+            exit(1)
+        else:
+            print(util.info('Successfully compiled ingen.'))
```
tonowak, line 51: This `if` looks like can be put into the `gen_util.compile_ingen` function. Is there any reason why error handling is here, instead of the function's definition?

src/sinol_make/commands/gen/__init__.py
```diff
@@ -0,0 +1,147 @@
...
+
+    def compile_correct_solution(self):
+        self.correct_solution_exe, compile_log_path = gen_util.compile_correct_solution(self.correct_solution, self.args,
+                                                                                        self.args.weak_compilation_flags)
+        if self.correct_solution_exe is None:
+            print(util.error('Failed ingen compilation.'))
```
tonowak, line 57: Should that indeed be `ingen`?

src/sinol_make/commands/gen/__init__.py
```diff
@@ -0,0 +1,147 @@
...
+                if result:
+                    print(util.info(f'Successfully generated output file {os.path.basename(arguments[i].output_test)}'))
+                else:
+                    failed = True
+                    print(util.error(f'Failed to generate output file {os.path.basename(arguments[i].output_test)}'))
+
+        if failed:
+            util.exit_with_error('Failed to generate some output files.')
+        else:
+            print(util.info('Successfully generated all output files.'))
```
tonowak, line 83: I'm not really a fan of the "creating a bool, changing it in a for statement, checking its results after the for" pattern, which shows up really often in the whole sinol-make's current codebase. How easy would it be to change it into a functional approach? E.g. by first creating the list of tests, then transforming it (through e.g. a `map` function, but I'm not sure whether it exists in Python) to a list of bools (that contain information whether the test was generated successfully), then checking whether a list contains only True bools (I remember a `all(list)` in Python, but there might be other ways) and then matching on the result (through a `if` or `match`)? Python might be not the best language for implementing this specific approach, but I think that there can be some other good approaches besides the current one (it's worth to spend some time googling this subject). I'm bringing this up, because it's easy to forget somewhere a variable assignment and it also increases the amount of variables in the environment, which increases the needed context to read the code, so I think this pattern is bug-prone.

src/sinol_make/commands/gen/__init__.py
```diff
@@ -0,0 +1,147 @@
...
+        """
+        old_md5_sums = None
+        if os.path.exists(os.path.join(os.getcwd(), 'in', '.md5sums')):
+            try:
+                with open(os.path.join(os.getcwd(), 'in', '.md5sums'), 'r') as f:
+                    old_md5_sums = yaml.load(f, Loader=yaml.FullLoader)
+            except yaml.YAMLError:
+                pass
+            if not isinstance(old_md5_sums, dict):
+                old_md5_sums = None
```
tonowak, line 98: This code's flow is a bit chaotic. Can it be made better?

src/sinol_make/commands/gen/__init__.py
```diff
@@ -0,0 +1,147 @@
...
+
+        solution_compiled = False
```
tonowak, line 100: Such booleans that are being changed in an already complicated block of code make it much harder to understand (and verify) the code. In such cases, I prefer to have a lambda function `compile_solution_when_needed` that isolates the caching logic, and then in the code I just use `compile_solution_when_needed` without any state/boolean logic that is not really important to the current block. Bonus points for completely hiding any variables and exposing only the lambda function (though I don't know how easy it is to make it in Python).

src/sinol_make/commands/gen/gen_util.py
```diff
@@ -0,0 +1,124 @@
...
+        else:
+            return None
+
+    if task_id is None:
+        task_id = '*'
+    ingen = glob.glob(os.path.join(os.getcwd(), 'prog', task_id + 'ingen.*'))
+    if len(ingen) == 0:
+        return None
+
+    # Makefiles prioritize non .sh ingens.
```
tonowak, line 35:
```suggestion
    # Sio2 first chooses shell scripts, then non-shell source codes.
```

src/sinol_make/commands/gen/gen_util.py
```diff
@@ -0,0 +1,124 @@
...
+    while process.poll() is None:
+        line = process.stdout.readline()
+        if line:
+            print(line.decode('utf-8'), end='')
+    line = process.stdout.read().decode('utf-8').strip()
+    if line:
+        print(process.stdout.read().decode('utf-8'), end='')
+    exit_code = process.returncode
+
+    return exit_code == 0
```
tonowak, line 92: Is there no better way to do it?

tests/packages/gen/prog/geningen.sh
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash
+
+# This script generates input tests in parallel.
+# Test names are read from file "tests" in the same directory split by spaces.
```
tonowak, line 4: Before settling on a specific ingen shell script which we will use, I first want to ask around. Maybe someone has some ready scripts, or maybe someone has some ideas how the script can be improved. The decision for merging this PR is up to you, I just want to postpone the review of this specific file.

tonowak (commented):
Looks good! Some slight suggestions to improve the code quality:

src/sinol_make/commands/gen/__init__.py
MasloMaslane: That this argument is optional and if present, only one string is taken

src/sinol_make/commands/gen/__init__.py
MasloMaslane: I moved it there

src/sinol_make/commands/gen/__init__.py
MasloMaslane: I changed it a bit, but I don't think it's really possible in this case, because `Pool.imap` executes functions in order, and returns result as soon as possible. If we completely removed printing in the for then nothing would print for a long time.

src/sinol_make/commands/gen/__init__.py
MasloMaslane: I moved the compilation away from this place bb39266fe800475f9b8fc8957dd84dc6823b608e

src/sinol_make/commands/gen/gen_util.py
MasloMaslane: I changed it a bit

tests/packages/gen/prog/geningen.sh
MasloMaslane: Sure, I think we should leave it for testing if the command works

src/sinol_make/commands/gen/__init__.py
MasloMaslane: I couldn't come up with anyting better

MasloMaslane: Maybe except for adding `md5_file = os.path.join(os.getcwd(), 'in', '.md5sums')`

src/sinol_make/commands/gen/__init__.py
tonowak: You can still print in the for. That can be done by mapping each result, printing a message when needed, and returning a bool. After that, you check whether all bools returned true through e.g. the `all` function. What do you think of it?

src/sinol_make/commands/gen/__init__.py
tonowak: Hmm, maybe checking whether the file exists and reading it can be combined into one?

tonowak:
```
old_md5_sums = None
try:
    with open(os.path.join(os.getcwd(), 'in', '.md5sums'), 'r') as f:
        if isinstance(old_md5_sums, dict):
            old_md5_sums = yaml.load(f, Loader=yaml.FullLoader)
except smth:
    pass
```
(hand-written, it might not work)

src/sinol_make/commands/gen/__init__.py
MasloMaslane: I fixed this

src/sinol_make/commands/gen/gen_util.py
```diff
@@ -0,0 +1,138 @@
...
+                    will return any ingen matching "*ingen.*"
+    :param ingen_path: path to ingen source file
+    :return: path to ingen source file or None if not found
+    """
+
+    if ingen_path is not None:
+        if os.path.exists(ingen_path):
+            return ingen_path
+        else:
+            return None
```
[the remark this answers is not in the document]
tonowak: I've checked for more places and indeed, there are *a lot* of `None` values in the code. That's not really good error handling. Usually when people prefer to return the errors (instead of throwing an exception), they use a kind of Error object that contains the result or an error message, which can be propagated up. That allows to contain all the logic in the helper functions, instead of cluttering the functions with the business logic and error messages. Though when possible, exiting deep in functions is also OK, as that get rids of the error handling and unnecessary code.

src/sinol_make/commands/gen/__init__.py
MasloMaslane: I changed it, now it looks similar to what you wrote

src/sinol_make/commands/gen/gen_util.py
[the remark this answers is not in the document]
MasloMaslane: I replaced the `None`s with errors in this command. I think that doing it for other commands should be in another PR.

src/sinol_make/commands/gen/gen_util.py
```diff
@@ -0,0 +1,138 @@
...
+    """
+    Compiles ingen and returns path to compiled executable.
+    """
+    compilers = compiler.verify_compilers(args, [ingen_path])
+    ingen_exe, compile_log_path = compile.compile_file(ingen_path, package_util.get_executable(ingen_path), compilers, weak_compilation_flags)
+
+    if ingen_exe is None:
+        print(util.error('Failed ingen compilation.'))
+        compile.print_compile_log(compile_log_path)
+        exit(1)
```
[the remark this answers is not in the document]
MasloMaslane: I added the optional arguemnt

src/sinol_make/commands/gen/gen_util.py
```diff
@@ -0,0 +1,135 @@
...
+
+def run_ingen(ingen_exe):
+    """
+    Runs ingen and generates all input files.
+    :param ingen_exe: path to ingen executable
+    :return: True if ingen was successful, False otherwise
+    """
+    is_shell = os.path.splitext(ingen_exe)[1] == '.sh'
+    if is_shell:
+        os.system(f'chmod +x {ingen_exe}')
```
tonowak, line 93: Does it work correctly on other operating system? Does e.g. Windows have a `chmod +x` command? Reminder that some operating systems can have different shell commands.

src/sinol_make/commands/gen/__init__.py
```diff
@@ -0,0 +1,119 @@
...
+                results.append(result)
+                if result:
+                    print(util.info(f'Successfully generated output file {os.path.basename(arguments[i].output_test)}'))
+                else:
+                    print(util.error(f'Failed to generate output file {os.path.basename(arguments[i].output_test)}'))
+
+        if not all(results):
+            util.exit_with_error('Failed to generate some output files.')
+        else:
+            print(util.info('Successfully generated all output files.'))
```
tonowak, line 63:
```suggestion
        with mp.Pool(self.args.cpus) as pool:
            results = []
            for i, result in enumerate(pool.imap(gen_util.generate_output, arguments)):
                results.append(result)
                if result:
                    print(util.info(f'Successfully generated output file {os.path.basename(arguments[i].output_test)}'))
                else:
                    print(util.error(f'Failed to generate output file {os.path.basename(arguments[i].output_test)}'))

            if not all(results):
                util.exit_with_error('Failed to generate some output files.')
            else:
                print(util.info('Successfully generated all output files.'))
```

tonowak: Reducing the scope of a variable greatly reduces the size of context needed to read the whole code.

src/sinol_make/commands/gen/__init__.py
```diff
@@ -0,0 +1,119 @@
...
+        """
+        Calculates md5 sums for each test.
+        :return: Tuple (dictionary of md5 sums, list of outputs tests that need to be generated)
+        """
+        old_md5_sums = None
+        try:
+            with open(os.path.join(os.getcwd(), 'in', '.md5sums'), 'r') as f:
+                old_md5_sums = yaml.load(f, Loader=yaml.FullLoader)
+            if not isinstance(old_md5_sums, dict):
+                old_md5_sums = None
```
tonowak, line 75:
```suggestion
            with open(os.path.join(os.getcwd(), 'in', '.md5sums'), 'r') as f:
                if isinstance(old_md5_sums, dict):
                    old_md5_sums = yaml.load(f, Loader=yaml.FullLoader)
```
Didn't I already make a similar suggestion previously? Is there a reason for not applying it?

tonowak (commented):
After those suggestions I'll approve.

src/sinol_make/commands/gen/__init__.py
MasloMaslane: Weird, I thought I changed it

MasloMaslane: Your code, won't work because `old_md5sums` is `None`, so the if will always fail.

Now the code looks like this:
```python
        old_md5_sums = None
        try:
            with open(os.path.join(os.getcwd(), 'in', '.md5sums'), 'r') as f:
                old_md5_sums = yaml.load(f, Loader=yaml.FullLoader)
                if not isinstance(old_md5_sums, dict):
                    old_md5_sums = None
        except (yaml.YAMLError, OSError):
            pass
```

tonowak: The flow of this code is still chaotic, it's hard to understand without any explanation or comments. Maybe first save the yaml to another variable with a clear name, and if its content is good, set the `old_md5_sums` to it?

tonowak: Yeah, I think it is now easier to understand the code's flow. And what do you think?

MasloMaslane: I agree that it's better than before

tonowak (approved):

PR #64, opened 2023-08-02
Add support for packages with library

MasloMaslane:


src/sinol_make/helpers/compile.py
```diff
@@ -1,29 +1,50 @@
...
+    if extra_compilation_files is None:
+        extra_compilation_files = []
+
+    for file in extra_compilation_files:
+        shutil.copy(file, os.path.join(os.path.dirname(output), os.path.basename(file)))
+
     gcc_compilation_flags = '-Werror -Wall -Wextra -Wshadow -Wconversion -Wno-unused-result -Wfloat-equal'
     if weak_compilation_flags:
-        gcc_compilation_flags = '-w' # Disable all warnings
+        gcc_compilation_flags = '-w'  # Disable all warnings
```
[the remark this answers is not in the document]
tonowak: Good to know, thanks!

MasloMaslane:
`makefile.in` requires a change, but I think that this should happen in `export` command

tonowak:
Sounds reasonable. Is there any way to reduce the number of files the user has to specify in the sinol-make config file?

MasloMaslane:
Not currently, but I guess sio2 could assume that `???lib.*` are library files and use them during compilation

tonowak:
You mean sio2 or sinol-make?

MasloMaslane:
Oh, you mean that `sinol-make` could detect library files and add them to `config.yml`?

tonowak:
Ahh, the `extra_compilation_*` variables are for sio2, not sinol-make. My bad.

src/sinol_make/commands/run/__init__.py
```diff
@@ -315,20 +315,33 @@ def compile_solutions(self, solutions):
...
+                for file in self.config["extra_compilation_args"].get(lang, []):
+                    extra_compilation_args.append(os.path.join(os.getcwd(), "prog", file))
+
+            for file in self.config.get("extra_compilation_files", []):
+                extra_compilation_files.append(os.path.join(os.getcwd(), "prog", file))
+
         try:
             compile.compile(source_file, output, self.compilers,
-                            open(compile_log_file, "w"), self.args.weak_compilation_flags)
+                            open(compile_log_file, "w"), self.args.weak_compilation_flags, extra_compilation_args,
```
tonowak, line 343: By the way, shouldn't that be a `with open`? It seems that this file descriptor is not being closed, right?

tonowak (commented):
Small nitpick, besides that LGTM.

src/sinol_make/commands/run/__init__.py
MasloMaslane: Yes, it should

src/sinol_make/helpers/compile.py
```diff
@@ -71,15 +94,14 @@ def compile_file(file_path: str, name: str, compilers: Compilers, weak_compilati
...
-            return output, compile_log_path
-        else:
+    with open(compile_log_path, 'w') as compile_log:
+        try:
+            if compile(file_path, output, compilers, compile_log, weak_compilation_flags):
+                return output, compile_log_path
+            else:
+                return None, compile_log_path
+        except CompilationError:
             return None, compile_log_path
```
tonowak, line 104:
```suggestion
        except CompilationError:
            pass
        return None, compile_log_path
```
Maybe it can be further simplified, I'm not familiar with exceptions in Python.

src/sinol_make/helpers/compile.py
```diff
@@ -88,8 +110,8 @@ def print_compile_log(compile_log_path: str):
...
     """
 
-    compile_log = open(compile_log_path, 'r')
-    lines = compile_log.readlines()
-    compile_log.close()
-    for line in lines[:500]:
-        print(line, end='')
+    with open(compile_log_path, 'r') as compile_log:
+        lines = compile_log.readlines()
+        compile_log.close()
```
tonowak, line 115:
```suggestion
```

src/sinol_make/util.py
```diff
@@ -79,7 +79,8 @@ def install_oiejq():
         raise Exception('Couldn\'t download oiejq (https://oij.edu.pl/zawodnik/srodowisko/oiejq.tar.gz couldn\'t connect)')
     if request.status_code != 200:
         raise Exception('Couldn\'t download oiejq (https://oij.edu.pl/zawodnik/srodowisko/oiejq.tar.gz returned status code: ' + str(request.status_code) + ')')
-    open('/tmp/oiejq.tar.gz', 'wb').write(request.content)
+    with open('/tmp/oiejq.tar.gz', 'wb') as oiejq_file:
+        oiejq_file.write(request.content)
```
tonowak, line 83: Sometimes you're using a cache directory in the project, sometimes a `/tmp` directory. Let's be consistent.

tonowak: I think it's possible in Python to e.g. create a directory that gets removed after the code leaves a 'with' block. Might be helpful here, but I haven't put much thought into it.

MasloMaslane: I changed it

tonowak: We're still inconsistent -- here we use a `/tmp` directory, in other places we use the cache directory in the project.

MasloMaslane: I think that oiejq installation should be in the temp directory, as there is no guarantee that the current directory is a package, and I thought the cache dir is for stuff the package creator might want to look into.

tonowak: Sounds reasonable. You can write an explanation in a comment.

tonowak: In fact, anything that is different than the usual style should have a comment, as probably I'll nevertheless ask about it during code review.

MasloMaslane: I added the comment.

tonowak (approved):

PR #66, opened 2023-08-04
`export` command

MasloMaslane:
- [x] export to `.tgz` for oioioi upload
- [x] ~~export to `ocen` debian package~~
- [x] ~~export to `dlazaw` debian package~~


MasloMaslane:
There is a lot of template code for `ocen` package

MasloMaslane:
I moved the package information from `pyproject.toml` to `setup.cfg` because previously I couldn't find a way to install sinol-make with the templates directory

src/sinol_make/commands/export/__init__.py
```diff
@@ -0,0 +1,287 @@
...
+    def get_name(self):
+        return "export"
+
+    def configure_subparser(self, subparser: argparse.ArgumentParser):
+        parser = subparser.add_parser(
+            self.get_name(),
+            help='Create archive for oioioi upload',
+            description='Creates archive ready to upload to oioioi. '
+                        'It is possible to create ocen and dlazaw debian packages for usage in contests '
+                        'with --ocen flag.'
```
tonowak, line 29: Why don't we always create the ocen package? Is there a reason not to?

src/sinol_make/commands/export/__init__.py
```diff
@@ -0,0 +1,287 @@
...
+            ingen_path = gen_util.get_ingen(self.task_id)
+            if os.path.splitext(ingen_path)[1] == '.sh':
+                ingen_exe = ingen_path
+            else:
+                ingen_exe = gen_util.compile_ingen(ingen_path, self.args, self.args.weak_compilation_flags)
+            if not gen_util.run_ingen(ingen_exe, tmpdir):
+                util.exit_with_error('Failed to run ingen.')
+
+            tests = glob.glob(os.path.join(tmpdir, f'{self.task_id}*.in'))
+            return [package_util.extract_test_id(test) for test in tests]
```
tonowak, line 54: So, I wonder, do we have to run the ingen in extract? While it's true that the old Makefile scripts did so, maybe there is a better way. Why not just taking all `.in` files? Is it a good idea, or a bad idea?

src/sinol_make/commands/export/__init__.py
```diff
@@ -0,0 +1,287 @@
...
+                    shutil.copy(file_path, target_dir)
+
+        print('Generating tests...')
+        generated_tests = self.get_generated_tests()
+        print('Copying only handwritten tests...')
+        for ext in ['in', 'out']:
+            os.mkdir(os.path.join(target_dir, ext))
+            for test in glob.glob(os.path.join(os.getcwd(), ext, f'{self.task_id}*.{ext}')):
+                if package_util.extract_test_id(test) not in generated_tests:
+                    shutil.copy(test, os.path.join(target_dir, ext))
```
tonowak, line 78: I think that tests `abc0*.in` should always be copied, as often the latex task statement requires them to exist to compile (because e.g. of the sample input/outputs in the statement). And also, handwritten tests without ingen might be a bit of a bad practice, we should discuss whether we want to discourage it or whether it is fine. What do you think?

src/sinol_make/commands/export/__init__.py
```diff
@@ -0,0 +1,287 @@
...
+
+    def ocen_gen_conf(self, ocen_dir, nums):
+        """
+        Generates ocen configuration file.
+        :param ocen_dir: Ocen directory path.
+        :param nums: Set of test numbers.
+        """
+
+        print("Generating ocen configuration file...")
+        with open(os.path.join(ocen_dir, 'oi.conf'), "w") as conf:
```
tonowak, line 139: Oh wow, this file exists and sio2 uses it? I didn't know it.

src/sinol_make/commands/export/__init__.py
```diff
@@ -0,0 +1,287 @@
...
+        def replace_text(file, old, new):
+            with open(file, 'r') as f:
+                content = f.read().replace(old, new)
+            with open(file, 'w') as f:
+                f.write(content)
+
+        replace_text(os.path.join(deb_dir, 'DEBIAN', 'control'), 'TASK_ID', self.task_id)
+        replace_text(os.path.join(deb_dir, 'DEBIAN', 'postinst'), 'task_id = None', f'task_id = "{self.task_id}"')
+
+        return self.build_debian_package(tmpdir, f'oi-ocen-{self.task_id}')
```
tonowak, line 199: I have absolutely no idea what is going on here. I'll re-read this code after getting to understand the debian packages.

src/sinol_make/commands/export/__init__.py
```diff
@@ -0,0 +1,287 @@
...
+        shutil.copytree(os.path.join(os.getcwd(), 'dlazaw'),
+                        os.path.join(tmpdir, 'oi-dlazaw', 'home', 'zawodnik', 'dlazaw'))
+
+        print("Creating debian package...")
+        return self.build_debian_package(tmpdir, 'oi-dlazaw')
+
+    def run(self, args: argparse.Namespace):
+        if not util.check_if_project():
+            print(util.warning('You are not in a project directory (couldn\'t find config.yml in current directory).'))
+            exit(1)
```
tonowak, line 255: Is this code copy-pasted from other commands? It seems reasonable to create a helper function for that.

src/sinol_make/templates/ocen-linux-sio2jail/LICENSE.txt
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2018 Wojciech Dubiel, Tadeusz Dudkiewicz, Przemysław Jakub Kozłowski, Maciej Wachulec
```
tonowak, line 3: The formatting of polish letters is weird here.

tonowak (commented):
Nice work, I'm very excited to merge this! This command will save us a lot of time before the competitions and hopefully get rid of problems we had during the competitions :D

src/sinol_make/commands/export/__init__.py
MasloMaslane: This requires an additional tool `dpkg` because the ocen packages are .deb files. And also only OI and OIJ use ocen packages

src/sinol_make/commands/export/__init__.py
MasloMaslane: There is a size limit for uploading packages, so I think it’s quite important

src/sinol_make/commands/export/__init__.py
```diff
@@ -0,0 +1,287 @@
...
+            tests = glob.glob(os.path.join(tmpdir, f'{self.task_id}*.in'))
+            return [package_util.extract_test_id(test) for test in tests]
+
+    def copy_files(self, target_dir: str):
+        """
+        Copies files from current directory to target directory.
+        :param target_dir: Directory to copy files to.
+        """
+        files = ['config.yml', 'makefile.in', 'Makefile.in',
+                 'prog', 'doc', 'public', 'dlazaw']
```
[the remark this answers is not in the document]
MasloMaslane: I found it when I was reading sio’s package processing code, files in this directory will be visible in “Files” tab in the contest. But I haven’t tested if this actually works

src/sinol_make/commands/export/__init__.py
MasloMaslane: No, this files is used by the “ocen” script on the contestants’ computers. It tells the script what tasks and tests there are. The constants are also hard cored in our original scripts for creating ocen packages

MasloMaslane: The constants can be changed by the contestants

src/sinol_make/commands/export/__init__.py
```diff
@@ -0,0 +1,287 @@
...
+        with open(os.path.join(ocen_dir, 'oi.conf'), "w") as conf:
+            conf.write(f'TASKS="{self.task_id}"\n'
+                       f'TESTS_{self.task_id}="{" ".join(sorted(nums))}"\n'
+                       f'HARD_TIMELIMIT=60    # limit czasu na wykonanie pod oitimetool (zwroc  uwage,\n'
+                       f'                     # ze nie ma to nic wspolnego z wynikiem oitimetool dla Twojego programu)\n'
+                       f'MEMLIMIT=1000000   # limit wykorzysztania pamieci przez program oraz oitimetool (KiB)\n'
+                       f'SRCLIMIT=100     # limit na rozmiar pliku zrodlowego (KiB)\n'
+                       f'EXELIMIT=10240   # limit na rozmiar pliku wykonywalnego (KiB) = 10MB\n')
+
+    def build_debian_package(self, package_path, package_dir_name):
```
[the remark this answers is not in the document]
MasloMaslane: Ocen and dlazaw packages are Debian packages. It makes it easier for us to install them on the contestants’ computers

src/sinol_make/commands/export/__init__.py
MasloMaslane: All Debian packages have a “debian” directory with various package information and installation scripts inside, and everything outside this folder will be placed into root of the file system in the same directories as they are in the package. For example a file which is in usr/share/file in the package will be place into /usr/share/file in the file system.

So first the code copies the package template with necessary files in “debian” directory, creates archive with the ocen directory and places it in the usr/share/oi-ocen-{task_id}/. On installation the archive is unpacked and placed into the proper place in the home directory.

setup.cfg
```diff
@@ -0,0 +1,50 @@
...
+    Homepage = https://github.com/sio2project/sinol-make
+classifiers =
+    Programming Language :: Python :: 3
+    License :: OSI Approved :: MIT License
+    Operating System :: OS Independent
+
+[options]
+packages = find_namespace:
+packages_dir =
+    = src
```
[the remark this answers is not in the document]
MasloMaslane: It should

src/sinol_make/commands/export/__init__.py
tonowak: Ah, you mean to *not* put tests that are generated?

src/sinol_make/commands/export/__init__.py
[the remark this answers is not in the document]
tonowak: Let's test it, just in case. Maybe we can just put everything into `public`, instead of `dlazaw`?

src/sinol_make/commands/export/__init__.py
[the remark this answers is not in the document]
tonowak: I see. So it's not for the jury, just for the technical team?

src/sinol_make/commands/export/__init__.py
tonowak: I think that should be a `--dpkg` flag and in the description we should write that it is for the technical team.

src/sinol_make/commands/export/__init__.py
MasloMaslane: Yes

tonowak: Hmm, what about requiring from the user to automatically generate each input file? Then also regenerating tests would be easier, as we can just remove each test.

MasloMaslane: But what about a task with multiple correct answers to one input and in the task statement there is an explanation to a hand written output? It would also break many existing packages as there are some that don't even have ingen.

src/sinol_make/commands/export/__init__.py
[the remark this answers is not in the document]
MasloMaslane: `dlazaw` is also always uploaded to contestants' computers so that they can test it locally

src/sinol_make/commands/export/__init__.py
```diff
@@ -0,0 +1,287 @@
...
+        """
+        files = ['config.yml', 'makefile.in', 'Makefile.in',
+                 'prog', 'doc', 'public', 'dlazaw']
+        for file in files:
+            file_path = os.path.join(os.getcwd(), file)
+            if os.path.exists(file_path):
+                if os.path.isdir(file_path):
+                    shutil.copytree(file_path, os.path.join(target_dir, file))
+                else:
+                    shutil.copy(file_path, target_dir)
```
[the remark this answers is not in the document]
MasloMaslane: `shutil.copytree` is only for copying directories

src/sinol_make/commands/export/__init__.py
MasloMaslane: I just saw a OIJ package with all tests generated and no ingen. I don't have great knowledge about creating tasks, but I think there are cases that handwritten tests may be helpful

MasloMaslane: For example sometimes tests can be written during the contest for specific bad solutions

src/sinol_make/commands/export/__init__.py
```diff
@@ -0,0 +1,287 @@
...
+        Executes ingen to check what tests are generated.
+        """
+        with tempfile.TemporaryDirectory() as tmpdir:
+            ingen_path = gen_util.get_ingen(self.task_id)
+            if os.path.splitext(ingen_path)[1] == '.sh':
+                ingen_exe = ingen_path
+            else:
+                ingen_exe = gen_util.compile_ingen(ingen_path, self.args, self.args.weak_compilation_flags)
+            if not gen_util.run_ingen(ingen_exe, tmpdir):
+                util.exit_with_error('Failed to run ingen.')
```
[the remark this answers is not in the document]
MasloMaslane: I moved the check into `compile_ingen`

src/sinol_make/commands/export/__init__.py
```diff
@@ -0,0 +1,287 @@
...
+
+            tests = glob.glob(os.path.join(tmpdir, f'{self.task_id}*.in'))
+            return [package_util.extract_test_id(test) for test in tests]
+
+    def copy_files(self, target_dir: str):
+        """
+        Copies files from current directory to target directory.
+        :param target_dir: Directory to copy files to.
+        """
```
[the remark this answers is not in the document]
MasloMaslane: I renamed the function

src/sinol_make/commands/export/__init__.py
```diff
@@ -0,0 +1,287 @@
...
+            for test in glob.glob(os.path.join(os.getcwd(), ext, f'{self.task_id}*.{ext}')):
+                if package_util.extract_test_id(test) not in generated_tests:
+                    shutil.copy(test, os.path.join(target_dir, ext))
+
+    def create_files(self, target_dir: str, config: dict):
+        """
+        Creates required files in target directory (makefile.in).
+        :param target_dir: Directory to create files in.
+        :param config: Config dictionary.
+        """
```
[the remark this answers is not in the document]
MasloMaslane: I renamed the function

src/sinol_make/commands/export/__init__.py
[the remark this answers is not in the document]
MasloMaslane: I tested it and the correct dir name is `attachments`, and this actually works

src/sinol_make/commands/export/__init__.py
MasloMaslane: I added a helper function like that

src/sinol_make/commands/export/__init__.py
```diff
@@ -0,0 +1,287 @@
...
+        parser = subparser.add_parser(
+            self.get_name(),
+            help='Create archive for oioioi upload',
+            description='Creates archive ready to upload to oioioi. '
+                        'It is possible to create ocen and dlazaw debian packages for usage in contests '
+                        'with --ocen flag.'
+        )
+
+        parser.add_argument('-o', '--output', type=str, default='export',
+                            help='output directory. Default: export')
```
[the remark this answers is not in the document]
tonowak: Is there any situation where an user might want to specify the directory? Maybe specifying the filename makes a bit more sense? Though, I think both are not needed.

What do you think of creating the package in the top-level directory, instead of the export/ subdirectory?

src/sinol_make/commands/export/__init__.py
tonowak: > But what about a task with multiple correct answers to one input and in the task statement there is an explanation to a hand written output? 

An usual solution is to make an `if` in the model solution for this specific test and print the wanted output.

> It would also break many existing packages as there are some that don't even have ingen.

Fair enough. Let's stick with that. Though I think it should print a warning for any tests it copies that is hand-written, as maybe it is a test left out and it shouldn't be there.

src/sinol_make/commands/export/__init__.py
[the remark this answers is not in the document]
tonowak: So, can we fit everything into one directory, or must there be two separate ones?

src/sinol_make/commands/export/__init__.py
tonowak: Sounds good.

README.md
```diff
@@ -68,6 +68,8 @@ Whenever the new input differs from the previous one, the model solution will be
 You can also specify your ingen source file which will be used. Run `sinol-make gen --help` to see available flags.
 - `sinol-make inwer` -- Verifies whether input files are correct using your "inwer.cpp" program. You can specify what inwer
 program to use, what tests to check and how many CPUs to use. Run `sinol-make inwer --help` to see available flags.
+- `sinol-make export` -- Creates archive ready to upload to oioioi.  Output directory can be specified with the --output
+flag. Run `sinol-make export --help` to see all available flags.
```
tonowak, line 72:
```suggestion
- `sinol-make export` -- Creates archive ready to upload to sio2. Output directory can be specified with the --output
flag. Run `sinol-make export --help` to see all available flags.
```

setup.cfg
```diff
@@ -0,0 +1,49 @@
+[metadata]
+name = sinol_make
+version = attr: sinol_make.__version__
+author = Mateusz Masiarz
+author_email = m.masiarz@fri.edu.pl
+maintainer = Tomasz Nowak
+maintainer_email = tomasz.nowak@tonowak.com
```
tonowak, line 7: Actually, please also write yourself as a mainainter.

src/sinol_make/commands/export/__init__.py
```diff
@@ -0,0 +1,140 @@
...
+        util.exit_if_not_package()
+
+        self.args = args
+        export_dir = args.output or 'export'
+        self.task_id = package_util.get_task_id()
+        self.export_dir = os.path.abspath(export_dir)
+        if not os.path.exists(self.export_dir):
+            os.makedirs(self.export_dir)
+
+        with tempfile.TemporaryDirectory() as tmpdir:
```
tonowak, line 128: Did we settle the "cache vs tempdir" issue? I would prefer to be consistent in the use of those two directories.

MasloMaslane: Yes, I changed it to use `cache` dir for exporting

src/sinol_make/commands/export/__init__.py
MasloMaslane: I added a warning for this

src/sinol_make/commands/export/__init__.py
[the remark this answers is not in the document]
MasloMaslane: This would require task creators to zip `dlazaw` before placing it in the `attachments` directory. I don't think that this would save time for anyone and it would be bad for git.

README.md
MasloMaslane: I changed it, and also added szkopul here

tonowak (approved):
Can you ask Radoszewski on Zulip (on a chat with me) to test this functionality (after those changes will be available in pip)? Maybe he'll find some bugs or find ways to improve the command.

PR #67, opened 2023-08-05
Fixes for python 3.7 and 3.8

MasloMaslane:
There were some problems with types and `importlib.resources` in python 3.7 and 3.8.

### DIFF
--- a/src/sinol_make/util.py
+++ b/src/sinol_make/util.py
@@ -171,17 +171,27 @@ def save_config(config):
             yaml.dump(config, config_file)
 
 
-def check_for_updates(current_version) -> Union[str, None]:
+def import_importlib_resources():
     """
-    Function to check if there is a new version of sinol-make.
-    :param current_version: current version of sinol-make
-    :return: returns new version if there is one, None otherwise
+    Function to import importlib_resources.
+    For Python 3.8 and below, we use importlib_resources.
+    For Python 3.9 and above, we use importlib.resources.
     """
     python_version = sys.version_info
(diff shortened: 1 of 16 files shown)


tonowak:
Did you add any tests that checks whether the problem is indeed solved?

MasloMaslane:
I ran them locally on both 3.7 and 3.8. Do you think that we should run all tests on 3.7 and 3.11? This would make the tests even longer

tonowak:
If we can run in parallel, why not?

tonowak:
Is it better to have users for which the code doesn't work, or wait a bit with merging the PRs? :)

MasloMaslane:
Our local runner can't run more than one job at once, but I'll try adding more and see if they still work

tonowak:
Probably running on the oldest and latest Python version which we support will be enough.

MasloMaslane:
I added a new runner and now the tests are on 3.7 and 3.11

src/sinol_make/util.py
```diff
@@ -204,6 +210,12 @@ def check_version():
     Function that asynchronously checks for new version of sinol-make.
     Writes the newest version to data/version file.
     """
+    python_version = sys.version_info
+    if python_version.minor <= 8:
+        import importlib_resources as importlib
+    else:
+        import importlib.resources as importlib
+
```
tonowak, line 218: code duplication?

MasloMaslane: I added a function for importing the library

tonowak (approved):

PR #68, opened 2023-08-05
Fix running with `time_limits` and `memory_limits` present in config

MasloMaslane:
I kinda forgot that not all tests have to have the same time and memory limits. This PR fixes that.

### DIFF
--- a/src/sinol_make/commands/run/__init__.py
+++ b/src/sinol_make/commands/run/__init__.py
@@ -54,7 +54,7 @@ def update_group_status(group_status, new_status):
 
 
 def print_view(term_width, term_height, program_groups_scores, all_results, print_data: PrintData, names, executions,
-               groups, scores, tests, possible_score, time_limit, memory_limit, cpus, hide_memory):
+               groups, scores, tests, possible_score, cpus, hide_memory, config):
     width = term_width - 13  # First column has 6 characters, the " | " separator has 3 characters and 4 for margin
     programs_in_row = width // 13  # Each program has 10 characters and the " | " separator has 3 characters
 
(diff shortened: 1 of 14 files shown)


src/sinol_make/commands/run/__init__.py
```diff
@@ -63,9 +63,14 @@ def print_view(term_width, term_height, program_groups_scores, all_results, prin
     sys.stdout = output
 
     program_scores = collections.defaultdict(int)
-    program_times = collections.defaultdict(lambda: -1)
-    program_memory = collections.defaultdict(lambda: -1)
-    time_remaining = (len(executions) - print_data.i - 1) * 2 * time_limit / cpus / 1000.0
+    program_times = collections.defaultdict(lambda: (-1, 0))
```
tonowak, line 66: What is `(-1, 0)`? I don't see any definitions here, so it's hard to confirm the code.

src/sinol_make/helpers/package_util.py
```diff
@@ -49,3 +49,37 @@ def get_executable_path(solution: str) -> str:
...
+    test_group = get_group(test_path)
+
+    if "time_limits" in config:
+        if test_id in config["time_limits"]:
+            return config["time_limits"][test_id]
+        elif int(test_group) in config["time_limits"]:
+            return config["time_limits"][int(test_group)]
+        elif str(test_group) in config["time_limits"]:
+            return config["time_limits"][str(test_group)]
+    return config["time_limit"]
```
tonowak, line 68: That looks weird, is there really no better way to write this code?

src/sinol_make/commands/run/__init__.py
MasloMaslane: I added a comment describing it

src/sinol_make/helpers/package_util.py
MasloMaslane: I changed it a bit

tonowak (approved):

PR #83, opened 2023-08-06
Print message on `sinol-make` crash

### LINKED ISSUE #78: On crash show message where to submit issues

MasloMaslane:
Closes #78

### DIFF
--- a/src/sinol_make/__init__.py
+++ b/src/sinol_make/__init__.py
@@ -32,7 +32,7 @@ def configure_parsers():
     return parser
 
 
-def _main():
+def _main_exn():
     parser = configure_parsers()
     args = parser.parse_args()
     commands = util.get_commands()
@@ -67,7 +67,7 @@ def _main():
 
 def main():
     try:
-        _main()
+        _main_exn()
     except:
         print(traceback.format_exc())
         util.exit_with_error('An error occurred while running the command.\n'
--- a/src/sinol_make/__init__.py
+++ b/src/sinol_make/__init__.py
@@ -1,8 +1,10 @@
 # PYTHON_ARCOMPLETE_OK
 
 import argcomplete
+import traceback
 import argparse
 import sys
+import os
 
 from sinol_make import util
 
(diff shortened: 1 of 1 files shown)


tonowak:
How does an exception now looks like for the user?

tonowak:
Yeah, we don't really want to hide the reason from the user. I'm asking because the output format affects how the user might perceive the custom error message. Soon I'll try to improve it.

src/sinol_make/__init__.py
```diff
@@ -56,7 +59,16 @@ def main():
...
     parser.print_help()
+
+
+def main():
+    try:
+        _main()
+    except:
+        print(traceback.format_exc())
+        util.exit_with_error('An error occurred while running the command.\n'
+                             'You can submit an issue here: https://github.com/sio2project/sinol-make/issues')
```
tonowak, line 74:
```suggestion
        util.exit_with_error('An error occurred while running the command.\n'
                             'If that is a bug, please report it or submit a bugfix: https://github.com/sio2project/sinol-make/#reporting-bugs-and-contributing-code')
```

src/sinol_make/__init__.py
```diff
@@ -30,7 +32,7 @@ def configure_parsers():
     return parser
 
 
-def main():
+def _main():
```
tonowak, line 35:
```suggestion
def main_exn():
```

Is there any Python conventions to indicate that a function might throw? I'm familiar with the `_exn` suffix in other languages, but I'm not sure how it is in Python.

MasloMaslane: I didn’t find any, but I think we should add it here

src/sinol_make/__init__.py
```diff
@@ -30,7 +32,7 @@ def configure_parsers():
     return parser
 
 
-def main():
+def _main_exn():
```
tonowak, line 35:
```suggestion
def main_exn():
```
I don't see any reason to hide this function now that its name describes what it does.

tonowak (approved):
You can merge after resolving this comment

PR #85, opened 2023-08-06
Various `run` command fixes

### LINKED ISSUE #74: In `run` command don't show 0 points when whole group hasn't finished
When not all tests in group finished show underscore instead of 0 (for example _/20 instead of 0/20)

### LINKED ISSUE #81: Partial points don't work in `run`
Partial points don't sum up properly in the table summary

MasloMaslane:
- When a group hasn't finished show ??/20 points instead of 0/20 - closes #74 
- Fix summing of partial points - closes #81 
- Now ocen tests are in group 0
- Packages only with `time_limits` and no `time_limit` in config are now working. Similar with memory.

### DIFF
--- a/src/sinol_make/commands/run/__init__.py
+++ b/src/sinol_make/commands/run/__init__.py
@@ -141,7 +141,7 @@ def print_view(term_width, term_height, program_groups_scores, all_results, prin
                         util.color_red(group_status)),
                           "%3s/%3s" % (points, scores[group]),
                           end=" | ")
-                program_scores[program] += scores[group] if group_status == "OK" else 0
+                program_scores[program] += points if group_status == "OK" else 0
                 program_groups_scores[program][group] = {"status": group_status, "points": points}
             print()
         print(8 * " ", end=" | ")
--- a/src/sinol_make/commands/run/__init__.py
+++ b/src/sinol_make/commands/run/__init__.py
@@ -44,7 +44,7 @@ def color_time(time, limit):
 
 def colorize_status(status):
     if status == Status.OK: return util.bold(util.color_green(status))
-    if status == Status.NO_STATUS: return util.warning(status)
+    if status == Status.PENDING: return util.warning(status)
     return util.error(status)
 
 
@@ -125,14 +125,14 @@ def print_view(term_width, term_height, program_groups_scores, all_results, prin
                     elif status == Status.ML:
                         program_memory[program] = (2 * package_util.get_memory_limit(test, config, args),
                                                    package_util.get_memory_limit(test, config, args))
-                    if status == Status.NO_STATUS:
-                        group_status = Status.NO_STATUS
+                    if status == Status.PENDING:
+                        group_status = Status.PENDING
                         min_points = 0
                     else:
                         group_status = update_group_status(group_status, status)
 
                 points = math.ceil(min_points / 100 * scores[group])
-                if any([results[test].Status == Status.NO_STATUS for test in results]):
+                if any([results[test].Status == Status.PENDING for test in results]):
                     print(" " * 3 + ("?" * len(str(scores[group]))).rjust(3) +
                           f'/{str(scores[group]).rjust(3)}', end=' | ')
                 else:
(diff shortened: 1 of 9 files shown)


src/sinol_make/commands/run/__init__.py
```diff
@@ -127,15 +128,20 @@ def print_table_end():
...
                         group_status = update_group_status(group_status, status)
 
-                points = math.floor(min_points / 100 * scores[group])
-                print("%3s" % util.bold(util.color_green(group_status)) if group_status == "OK" else util.bold(
-                    util.color_red(group_status)),
-                      "%3s/%3s" % (points, scores[group]),
-                      end=" | ")
-                program_scores[program] += scores[group] if group_status == "OK" else 0
+                points = math.ceil(min_points / 100 * scores[group])
+                if not all_tests_finished:
```
tonowak, line 136:
```suggestion
                all_tests_finished = all([results[test].Status != "  " for test in results])
                if not all_tests_finished:
```
The "create a bool, update in in a for, use it further down" is a code pattern that reduces readability and increases chances of a bug (there are three places where a bug could happen, instead of one). Please try to reduce the amount of variables with wide scope in the code.

The suggestion was hand-written, it might not work.

The `status == "  "` is also not readable and seems like an easy place to make a bug. Is there any way to make it clearer?

tonowak: After improving the `status == "  "` case, the meaning of the `all(...)` statement would be so clear that a variable might not needed and just an `if not all(...)` would be perfectly readable and shorter.

tonowak (commented):
Nice, those fixes turned out to be simple, good to see.

src/sinol_make/commands/run/__init__.py
MasloMaslane: I changed it a bit

src/sinol_make/commands/run/__init__.py
```diff
@@ -79,6 +79,7 @@ def print_view(term_width, term_height, program_groups_scores, all_results, prin
     title = 'Done %4d/%4d. Time remaining (in the worst case): %5d seconds.' \
             % (print_data.i + 1, len(executions), time_remaining)
     title = title.center(term_width)
+    NO_STATUS = "  "
```
tonowak, line 82: Then why not also have similar "enums" for other status?

MasloMaslane: I added an enum for statuses

MasloMaslane:
I also added a new fix, now packages with `time_limits` and no `time_limit` in config work properly (and also with `memory_limits` and no `memory_limit`)

src/sinol_make/commands/run/__init__.py
```diff
@@ -42,20 +43,20 @@ def color_time(time, limit):
 
 
 def colorize_status(status):
-    if status == "OK": return util.bold(util.color_green(status))
-    if status == "  " or status == "??": return util.warning(status)
+    if status == Status.OK: return util.bold(util.color_green(status))
+    if status == Status.NO_STATUS or status == "??": return util.warning(status)
```
tonowak, line 47: What do you think of replacing `Status.NO_STATUS` with `Status.PENDING`? What does `status == '??'` mean?

src/sinol_make/commands/run/__init__.py
```diff
@@ -666,9 +673,23 @@ def compile_and_run(self, solutions):
         names = solutions
         return self.run_solutions(compiled_commands, names, solutions)
 
+    def convert_Status_to_string(self, dictionary):
```
tonowak, line 676:
```suggestion
    def convert_status_to_string(self, dictionary):
```

src/sinol_make/commands/run/__init__.py
MasloMaslane: `PENDING` is a better name. The `??` is also present in original Kostka's code, but it wasn't used there either

tonowak (approved):
Good job! :100:

PR #86, opened 2023-08-06
Fix running `gen` on package with library

MasloMaslane:


src/sinol_make/helpers/compile.py
```diff
@@ -95,11 +96,21 @@ def compile_file(file_path: str, name: str, compilers: Compilers, weak_compilati
...
     os.makedirs(compile_log_dir, exist_ok=True)
 
+    with open(os.path.join(os.getcwd(), "config.yml"), "r") as config_file:
+        config = yaml.load(config_file, Loader=yaml.FullLoader)
+
+    extra_compilation_files = [os.path.join(os.getcwd(), "prog", file)
+                               for file in config.get("extra_compilation_files", [])]
+    extra_compilation_args = []
+    if 'extra_compilation_args' in config:
+        extra_compilation_args = config['extra_compilation_args'].get(os.path.splitext(name)[1], [])
```
tonowak, line 106: Assuming that `config` is a dictionary, I bet Python has a "get" or "get_with_default" function that returns the object under the given key or a default value when the key is not present. That would make the code much shorter.

MasloMaslane: I changed it

src/sinol_make/helpers/compile.py
```diff
@@ -95,11 +96,19 @@ def compile_file(file_path: str, name: str, compilers: Compilers, weak_compilati
     os.makedirs(executable_dir, exist_ok=True)
     os.makedirs(compile_log_dir, exist_ok=True)
 
+    with open(os.path.join(os.getcwd(), "config.yml"), "r") as config_file:
+        config = yaml.load(config_file, Loader=yaml.FullLoader)
+
+    extra_compilation_files = [os.path.join(os.getcwd(), "prog", file)
+                               for file in config.get("extra_compilation_files", [])]
+    extra_compilation_args = config.get('extra_compilation_args', {}).get(os.path.splitext(name)[1], [])
```
tonowak, line 104: We can make this change also in other places in the code. What do you think?

MasloMaslane: I agree that this is the better way. I changed it in other places, but couldn't find much

tonowak (approved):

PR #89, opened 2023-08-14
Fixes and small changes to `oiejq` installation

MasloMaslane:
The main change is that I didn't really like how oiejq was unpacked into `~/.local/bin`, so now it has it's own directory there


tonowak (changes requested):
That is a downgrade in functionality. Most Linux users have `~/.local/bin/` in `$PATH`, thanks to which they could just use the `oiejq` command. Now they cannot run `oiejq` as simply as previously.

tonowak:
I would say it's better to rename the `oiejq` directory to something else (e.g. `oiejq_old_dir`) and install it the binary as `oiejq`. Thanks to that, probably the user will gain a new functionality (running `oiejq` thanks to `$PATH`) and one's files will still be present.

src/sinol_make/util.py
```diff
@@ -90,33 +98,39 @@ def install_oiejq():
...
-                yield member
-
         with tarfile.open(oiejq_path) as tar:
-            tar.extractall(path=os.path.expanduser('~/.local/bin'), members=strip(tar))
-        os.rename(os.path.expanduser('~/.local/bin/oiejq.sh'), os.path.expanduser('~/.local/bin/oiejq'))
+            tar.extractall(path=tmpdir)
+
+        # If the user has a directory named `oiejq` in `~/.local/bin`,
+        # we rename the `oiejq.sh` executable to `oiejq_sinol-make`.
+        # Otherwise, we rename it to `oiejq`.
```
tonowak, line 106: Maybe it's better to check whether `oiejq` can be run, and otherwise move it? That is, maybe we don't have to check whether it is a directory or not?

MasloMaslane: If there exists a file named `oiejq` in `.local/bin/` and it is executable, then it will be used.
I don't think that moving user's files is good

tonowak: Then I would prefer to make an error with an understandable error message.

tonowak: Thanks to that, the user would do whatever he wants with his files, and we won't have to lengthen the code because of edge-cases.

MasloMaslane: Now the code fails before installing if `~/.local/bin/oiejq` already exists

MasloMaslane:
I added the suggestion from https://github.com/sio2project/sinol-make/pull/97#discussion_r1306561499 here, now only `oiejq` and `sio2jail` are copied to `~/.local/bin`

src/sinol_make/util.py
```diff
@@ -33,39 +34,44 @@ def check_if_project():
     return False
 
 
+def _check_if_oiejq_executable(path):
+    if not os.path.isfile(path):
+        return False
+    if not os.access(path, os.X_OK):
+        return False
```
tonowak, line 41: Is the first `if` needed?

src/sinol_make/util.py
```diff
@@ -33,39 +34,44 @@ def check_if_project():
...
+
+    try:
+        p = subprocess.Popen([path], shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        p.wait()
+        if p.returncode == 0:
+            return True
+        else:
+            return False
```
tonowak, line 49:
```suggestion
        return p.returncode == 0
```

src/sinol_make/util.py
```diff
@@ -74,6 +80,13 @@ def install_oiejq():
     if not os.path.exists(os.path.expanduser('~/.local/bin')):
         os.makedirs(os.path.expanduser('~/.local/bin'), exist_ok=True)
 
+    if os.path.exists(os.path.expanduser('~/.local/bin/oiejq')) and \
+            not _check_if_oiejq_executable(os.path.expanduser('~/.local/bin/oiejq')):
```
tonowak, line 84: Isn't it enough to check only the second condition?

MasloMaslane: It isn't, because `_check_if_oiejq_executable(path)` returns `False` if the `path` doesn't exist

tonowak (approved):
Looks good!

PR #90, opened 2023-08-14
Change behavior of expected scores when using `--tests` flag

MasloMaslane:
Now only groups with all their tests run will be shown in expected scores


src/sinol_make/commands/run/__init__.py
```diff
@@ -725,9 +748,21 @@ def convert_to_expected(results):
...
+                groups_to_remove = []
+                for group in new_expected_scores[solution]["expected"]:
+                    if group not in used_groups:
+                        groups_to_remove.append(group)
+                for group in groups_to_remove:
+                    del new_expected_scores[solution]["expected"][group]
+                if len(new_expected_scores[solution]["expected"]) == 0:
+                    solutions_to_delete.append(solution)
+            for solution in solutions_to_delete:
+                del new_expected_scores[solution]
```
tonowak, line 764: It's hard to tell what this code does at a first glance. Do you have some ideas how that can be improved?

tonowak (commented):
Another bugfix, nice! :lady_beetle:

src/sinol_make/commands/run/__init__.py
MasloMaslane: I added some comments, is it better now?

src/sinol_make/commands/run/__init__.py
```diff
@@ -725,9 +748,25 @@ def convert_to_expected(results):
                 for group in config_expected_scores[solution]["expected"]:
                     used_groups.add(group)
         else:
-            for solution in results.keys():
-                for group in results[solution].keys():
-                    used_groups.add(group)
+            used_groups = self.get_whole_groups()
+
+            # This removes groups with not all tests run from `new_expected_scores` based on `used_groups`.
```
tonowak, line 753: I don't understand this comment.

src/sinol_make/commands/run/__init__.py
```diff
@@ -725,9 +748,25 @@ def convert_to_expected(results):
...
+            # If there was a solution with all groups removed, it is also removed.
```
tonowak, line 754: In fact, I also don't understand this line. Does a "solution" have "groups"?

MasloMaslane: I changed the comments. What do you think now?

tonowak (approved):
Yeah, thanks, now I understand!

PR #91, opened 2023-08-14
Change `time` installation steps in README

MasloMaslane:
@tonowak could you test if `pacman -S time` installs `time` in /usr/bin/time correctly? I can't test it on my system at the moment

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,8 @@ As `oiejq` works only on Linux-based operating systems,
 Nevertheless `sinol-make` supports those operating systems,
 though there are additional installation steps required to use
 other tools for measuring time (which are non-deterministic and produce reports different from sio2):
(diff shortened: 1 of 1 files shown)


tonowak:
> @tonowak could you test if pacman -S time installs time in /usr/bin/time correctly? I can't test it on my system at the moment

What about testing it through a GitHub Action? Basically an integration test that checks whether the `--time-tool time` flag works on all operating systems.

MasloMaslane:
Oh yeah, I forgot that tests already install and use time

tonowak:
Does it mean that `pacman -S time` is not needed?

MasloMaslane:
It is needed if one wants to use `time` for measuring

tonowak (approved):

PR #92, opened 2023-08-15
OI-like point assignment

MasloMaslane:
I will add a test package and test this on real tasks next week.
The code for calculating points was copied from `oioioi`


src/sinol_make/contest_types/__init__.py
```diff
@@ -0,0 +1,19 @@
...
+    with open(os.path.join(os.getcwd(), "config.yml"), "r") as config_file:
+        config = yaml.load(config_file, Loader=yaml.FullLoader)
+        contest_type = config.get("sinol_contest_type", "default")
+
+    if contest_type == "default":
+        return DefaultContest()
+    elif contest_type == "OI":
+        return OIContest()
+    else:
+        raise UnknownContestType(f'Unknown contest type "{contest_type}"')
```
tonowak, line 19: We should create an example `config.yml` that also serves as a documentation. Basically a sample `config.yml` with a lot of comments that explains possible options and logic/behavior. What do you think?

src/sinol_make/contest_types/default.py
```diff
@@ -0,0 +1,17 @@
+from math import floor
+from typing import List
+
+from sinol_make.commands.run.structs import ExecutionResult
+
+
+class DefaultContest:
+    """
+    Default contest type.
+    """
```
tonowak, line 10: How can we make this comment more useful? Currently it almost copy-pastes the class name.

tonowak (commented):
Good change, it unblocks using sinol-make in OI competitions :tada: !

src/sinol_make/contest_types/oi.py
```diff
@@ -0,0 +1,17 @@
...
+from sinol_make.contest_types.default import DefaultContest
+
+
+class OIContest(DefaultContest):
+    """
+    Contest type for Polish Olympiad in Informatics.
+    """
+
+    def get_test_score(self, result: ExecutionResult, time_limit, memory_limit):
+        """Full score if took less than half of limit and then decreasing to 1"""
```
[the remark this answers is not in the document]
MasloMaslane: Yes, this when someone has a correct answer that one point is always guaranteed

src/sinol_make/contest_types/__init__.py
MasloMaslane: Yes, that’s a good idea

MasloMaslane: I added an example config

src/sinol_make/contest_types/default.py
MasloMaslane: I added some explanations

src/sinol_make/contest_types/oi.py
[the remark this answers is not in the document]
MasloMaslane: I changed the comment

examples/config.yml
```diff
@@ -0,0 +1,78 @@
...
+  4: 25
+
+# Time limit for all tests is defined in `time_limit` key.
+# More precise time limit for each group or test can be defined in `time_limits` key.
+# The more precise time limit has higher priority (first test, then group, then global time limit).
+# Time limit is in milliseconds.
+time_limit: 1000
+time_limits:
+  1: 2000
+  1a: 2500
```
tonowak, line 28: Hmm, I think we don't want the functionality to specify the time limit on a particular test. Now we have a PDF during the competition which describes what is the exact time limit on a given group. Setting different time limits in a group is against this idea.

tonowak: In short, I believe that setting time limit on a test is a bad practice and we shouldn't allow the user to do so.

examples/config.yml
```diff
@@ -0,0 +1,78 @@
...
+#               Points for a group are calculated based of the lowest number of points for a test in this group.
+# - `oi` - Points for a test are unchanged if the execution time is less or equal to the time limit.
+#          Otherwise, number of points decreases linearly to one depending on the execution time.
+#          Points for a group are calculated same as in `default` mode.
+sinol_contest_type: oi
+
+# sinol-make can check if the solutions run as expected when using `run` command.
+# Key `sinol_expected_scores` defines expected scores for each solution on each tests.
+# There should be no reason to change this key manually.
+# It is automatically generated and managed by sinol-make.
```
tonowak, line 78:
```suggestion
# It is automatically generated and managed by sinol-make.
sinol_expected_scores: {}
```

Let's also add the default value, so that from the start the user is reminded that this key should be present in the config file. (the default value might be different than what I wrote)

tonowak (commented):
The `examples/config.yml` is excellent, I love it!

examples/config.yml
MasloMaslane: I agree that this is a bad practice, but we can't just delete it. But I think we could write that there should be no reason for using this.

tonowak: Why we can't just delete it?

tonowak: It's not like we'll remove it from sio2, just from sinol-make.

MasloMaslane: Entirely from the code or just from the example config?

tonowak: I would say that it's safer for the users to have this feature removed entirely from sinol-make. It's also not really good to have functionality that is not documented (and we don't really want to document this functionality).

MasloMaslane: Now `sinol-make` exits when there is a time or memory limit for one test

tonowak (approved):

PR #95, opened 2023-08-24
Allow `extra_compilation_args` to be a string or an array

### LINKED ISSUE #93: Code breaks when `extra_compilation_args` is a string and not an array

MasloMaslane:
Closes #93


src/sinol_make/commands/export/__init__.py
```diff
@@ -92,11 +92,16 @@ def create_makefile_in(self, target_dir: str, config: dict):
...
+                    return [obj]
+                return obj
+
             if 'extra_compilation_args' in config:
                 if 'cpp' in config['extra_compilation_args']:
-                    cxx_flags += ' ' + ' '.join(config['extra_compilation_args']['cpp'])
+                    cxx_flags += ' ' + ' '.join(to_array(config['extra_compilation_args']['cpp']))
                 if 'c' in config['extra_compilation_args']:
-                    c_flags += ' ' + ' '.join(config['extra_compilation_args']['c'])
+                    c_flags += ' ' + ' '.join(to_array(config['extra_compilation_args']['c']))
```
tonowak, line 104: The `to_array` is a *very specific* function. I bet we can make a function whose purpose is much clearer:

```
def format_multiple_arguments(obj):
  if isinstance(obj, str):
    return str
  return ' '.join(obj)

cxx_flags += ' ' + format_multiple_arguments(config[...])
```
Shorter and more readable!

Maybe before you put your code for review, could you look over it and think *can this piece of code be made shorter?*

src/sinol_make/helpers/compile.py
```diff
@@ -25,6 +25,8 @@ def compile(program, output, compilers: Compilers = None, compile_log = None, we
     """
     if extra_compilation_args is None:
         extra_compilation_args = []
+    if isinstance(extra_compilation_args, str):
+        extra_compilation_args = [extra_compilation_args]
```
tonowak, line 29: Usually, after such modifications (which serve the purpose of unifying the types to a single one), we can make the code more readable by exactly saying what the type should be, e.g. by making an `assert(isinstance(extra_compilation_args, Array[string]))` (handwritten, might not work)

tonowak (commented):
Whew, that's a lot of tests for a single change in the code, nice!

src/sinol_make/commands/export/__init__.py
MasloMaslane: I changed it

src/sinol_make/helpers/compile.py
MasloMaslane: What do you think about this assert `assert isinstance(extra_compilation_args, list) and all(isinstance(arg, str) for arg in extra_compilation_args)`? It's quite long, but there no other way to check if the type is list of strings

tonowak: It's good enough, I guess.

tonowak (approved):

PR #97, opened 2023-08-24
detect whether `kernel.perf_event_paranoid` is set

MasloMaslane:


src/sinol_make/oiejq/__init__.py
```diff
@@ -0,0 +1,124 @@
...
+        process = subprocess.Popen([oiejq, python_executable, test_file], stdout=tmpfile, stderr=subprocess.DEVNULL)
+        process.wait()
+        tmpfile.seek(0)
+        output = tmpfile.read().decode('utf-8')
+        process.terminate()
+
+    if output != "Test string\n":
+        util.exit_with_error("You don't have permission to use perf counters required by `oiejq`. "
+                             "Please run `sudo sysctl kernel.perf_event_paranoid=-1` or "
+                             "add `kernel.perf_event_paranoid=-1` to `/etc/sysctl.conf` and reboot.")
```
tonowak, line 124: The "sysctl" command is *not* equivalent to the "add to /etc/sysctl.conf", right? The first one works only for the current shell session and the second one is permament? (I'm not sure).

MasloMaslane: The first works until the system is rebooted

tonowak:
```suggestion
        util.exit_with_error("You don't have permission to use perf counters required by `oiejq`. "
                             "Please run `sudo sysctl kernel.perf_event_paranoid=-1` or "
                             "add `kernel.perf_event_paranoid=-1` to `/etc/sysctl.conf` and reboot.")
```
I would prefer something that explains the situation a bit more:

```
To use the recommended tool for measuring time called oiejq, please:
- execute `sudo sysctl kernel.perf_event_paranoid=-1` to make oiejq work for the current system session,
- or add `kernel.perf_event_paranoid=-1` to `/etc/sysctl.conf` and reboot to permamently make oiejq work.
For more details, see [link or path to readme of oiejq that explains those commands].
```

Let's also manually split the longer printed messages, so that they fit in terminals of users with 80 character width.

As oiejq doesn't work on MacOS and Windows, what do you think of writing an error that the recommended tool for measuring time oiejq doesn't work on those systems and the user can specify another tool for measuring time through the `--time` command?

MasloMaslane: I changed the message, but in oiejq's and sio2jail's readme it isn't explained what this command does.

The message for `oiejq` on Windows and macOS is here:
https://github.com/sio2project/sinol-make/blob/aea45b8d275712178fd5de86695551f6132dd55d/src/sinol_make/commands/run/__init__.py#L919-L921
but I think that it should be change to something like:
```
`oiejq` is only available on Linux. You can specify another time tool with `--time` flag.
```

tonowak: Let's also explain that `oiejq` is the recommended way of measuring time. Maybe we could link the README or copy-paste parts of it to this error message.

src/sinol_make/oiejq/__init__.py
```diff
@@ -119,6 +119,10 @@ def check_perf_counters_enabled():
...
-        util.exit_with_error("You don't have permission to use perf counters required by `oiejq`. "
-                             "Please run `sudo sysctl kernel.perf_event_paranoid=-1` or "
-                             "add `kernel.perf_event_paranoid=-1` to `/etc/sysctl.conf` and reboot.")
+        util.exit_with_error("To use the recommended tool for measuring time called oiejq, please:\n"
+                             "- execute `sudo sysctl kernel.perf_event_paranoid=-1` to make oiejq work for\n"
+                             "  the current system session,\n"
+                             "- or add `kernel.perf_event_paranoid=-1` to `/etc/sysctl.conf`\n"
+                             "  and reboot to permanently make oiejq work.\n"
+                             "For more details, see https://github.com/sio2project/sio2jail#running or\n"
+                             "read `~/.local/bin/readme.txt`.")
```
tonowak: Is that indeed the correct path of the readme? I'm wondering whether it's bad that we unzip all the files into the `~/.local/bin/` directory, not only `oiejq` and `sio2jail` but also other stuff.

MasloMaslane: That's a good idea, I'll add it in #89

src/sinol_make/oiejq/__init__.py
MasloMaslane: I changed this message

src/sinol_make/commands/run/__init__.py
```diff
@@ -917,13 +918,19 @@ def validate_arguments(self, args):
         timetool_path = None
         if args.time_tool == 'oiejq':
             if sys.platform != 'linux':
-                util.exit_with_error('oiejq is only available on Linux.')
+                util.exit_with_error('As `oiejq` works only on Linux-based operating systems,\n'
+                                     'we do not recommend using operating systems such as Windows or macOS.\n'
+                                     'Nevertheless, if you want to use them, you can specify\n'
+                                     'another time tool with `--time-tool` flag.\n'
```
tonowak, line 924:
```suggestion
                                     'Nevertheless, you can still run sinol-make by speficying another way of measuring time\n'
                                     'through the `--time-tool` flag. See `sinol-make run --help` for more information about the flag.\n'
```

tonowak: Probably this suggestion would need to be split into one more newline.

src/sinol_make/oiejq/__init__.py
```diff
@@ -0,0 +1,129 @@
...
+
+    def check(path):
+        try:
+            p = subprocess.Popen([path], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+            p.wait()
+            p.kill()
+            if p.returncode == 0:
+                return True
+            else:
+                return False
```
tonowak, line 26:
```suggestion
            return p.returncode == 0
```

src/sinol_make/oiejq/__init__.py
```diff
@@ -0,0 +1,129 @@
...
+        process.terminate()
+
+    if output != "Test string\n":
+        util.exit_with_error("To use the recommended tool for measuring time called oiejq, please:\n"
+                             "- execute `sudo sysctl kernel.perf_event_paranoid=-1` to make oiejq work for\n"
+                             "  the current system session,\n"
+                             "- or add `kernel.perf_event_paranoid=-1` to `/etc/sysctl.conf`\n"
+                             "  and reboot to permanently make oiejq work.\n"
+                             "For more details, see https://github.com/sio2project/sio2jail#running or\n"
+                             "read `~/.local/bin/readme.txt`.\n"
```
tonowak, line 128: This now needs to be changed.

src/sinol_make/oiejq/__init__.py
```diff
@@ -0,0 +1,129 @@
...
+                             "Please keep in mind that oiejq doesn't work on macOS and Windows WSL.\n")
```
tonowak, line 129: Is this line relevant for the user? If the user won't see this message when using MacOS or Windows (because he'll see another message), we don't have to print it.

MasloMaslane: I added this with the idea in mind that the user could see it when using Windows WSL, but I just found a way to check if running under WSL, so maybe I'll add it in another PR and remove this line

tonowak: Sounds good.

src/sinol_make/oiejq/__init__.py
MasloMaslane: I think we should just leave the link to the sio2jail's readme, since the oiejq's readme is only in the archive

tonowak: Maybe there is a short sentence in the readme which we can copy-paste here?

tonowak: Anything that might explain to the user that this dangerously looking `sudo` command is in fact not dangerous.

MasloMaslane: There is only this in the readme about the command:
```
Być może, żeby narzędzie działało, będziesz musiał przed pierwszym
uruchomieniem wykonać:

    sudo sysctl kernel.perf_event_paranoid=-1

albo ustawić analogiczną zmienną w `/etc/sysctl.conf
```
so nothing helpful

tonowak: OK, let's not mention why and let's remove the path to readme.

src/sinol_make/commands/run/__init__.py
```diff
@@ -917,13 +918,19 @@ def validate_arguments(self, args):
         timetool_path = None
         if args.time_tool == 'oiejq':
             if sys.platform != 'linux':
-                util.exit_with_error('oiejq is only available on Linux.')
+                util.exit_with_error('As `oiejq` works only on Linux-based operating systems,\n'
+                                     'we do not recommend using operating systems such as Windows or macOS.\n'
+                                     'Nevertheless, you can still run sinol-make by specifying\n'
+                                     'another way of measuring time through the `--time-tool` flag.\n'
+                                     'See `sinol-make run --help` for more information about the flag.\n')
```
tonowak, line 925:
```suggestion
                                     'See `sinol-make run --help` for more information about the flag.\n
                                     'See https://github.com/sio2project/sinol-make#why for more information about `oiejq`.\n')
```

tonowak (approved):
Resolve this last comment before merging:

src/sinol_make/commands/run/__init__.py
```diff
@@ -917,13 +918,20 @@ def validate_arguments(self, args):
         timetool_path = None
         if args.time_tool == 'oiejq':
             if sys.platform != 'linux':
-                util.exit_with_error('oiejq is only available on Linux.')
+                util.exit_with_error('As `oiejq` works only on Linux-based operating systems,\n'
+                                     'we do not recommend using operating systems such as Windows or macOS.\n'
+                                     'Nevertheless, you can still run sinol-make by specifying\n'
+                                     'another way of measuring time through the `--time-tool` flag.\n'
+                                     'See `sinol-make run --help` for more information about the flag.\n
```
tonowak, line 925:
```suggestion
                                     'See `sinol-make run --help` for more information about the flag.\n'
```

PR #102, opened 2023-08-29
Add function to check if running under Windows WSL

MasloMaslane:


src/sinol_make/util.py
```diff
@@ -243,6 +244,20 @@ def stringify_keys(d):
         return d
 
 
+def is_wsl():
+    """
+    Function to check if the program is running on Windows Subsystem for Linux.
+    """
+    return "microsoft" in platform.uname().release.lower()
```
tonowak, line 251: Does this also return true when the user is running just the Windows shell, not the WSL?

MasloMaslane: I can't test it now, but I changed it now so that this detects only wsl

tonowak (approved):

PR #103, opened 2023-08-29
Change stack size to unlimited before running

MasloMaslane:


tonowak (commented):
If we can, we should set the stack size to a fixed amount, e.g. 1GB or 1.5GB. It avoids infinite recursion which sometimes hangs the PC.

tonowak:
To check whether that works, we can have a test with a recursive function that goes deep, e.g. 1e8 times deep.

MasloMaslane:
Maybe if we set stack size to the memory limit then it would behave the same as unlimited stack size? I'm not sure

tonowak:
Sounds good. Is the memory limit always set?

MasloMaslane:
It should be

src/sinol_make/util.py
```diff
@@ -244,14 +244,28 @@ def stringify_keys(d):
         return d
 
 
-def change_stack_size_to_unlimited():
+def change_stack_size():
     """
-    Function to change the stack size to unlimited.
+    Function to change the stack size to max memory limit.
     """
+    max_memory = 60000 * 1024  # Default memory limit is 66000kB, `resource` module uses bytes.
```
tonowak, line 251:
```suggestion
    default_max_memory = 60000 * 1024  # Default memory limit is 66000kB, `resource` module uses bytes.
```

src/sinol_make/util.py
```diff
@@ -244,14 +244,28 @@ def stringify_keys(d):
...
+        config = yaml.load(config_file, Loader=yaml.FullLoader)
+
+    max_memory = max(config.get("memory_limit", 0) * 1024, max_memory)
+    for memory_limit in config.get("memory_limits", {}).values():
+        max_memory = max(memory_limit * 1024, max_memory)
+    for override in config.get("override_limits", {}).values():
+        max_memory = max(override.get("memory_limit", 0) * 1024, max_memory)
+        for memory_limit in override.get("memory_limits", {}).values():
+            max_memory = max(memory_limit * 1024, max_memory)
+    hard_limit = resource.getrlimit(resource.RLIMIT_STACK)[1]
```
tonowak, line 262: Any ideas how to shorten the code, or split the functionality into different sections?

Also, I'm not so sure that we care about the "max memory out of all possible options". The `max` function seems not necessarily correct. What do you think?

src/sinol_make/util.py
```diff
@@ -244,14 +244,28 @@ def stringify_keys(d):
...
+
     try:
-        resource.setrlimit(resource.RLIMIT_STACK, (resource.RLIM_INFINITY, resource.RLIM_INFINITY))
+        resource.setrlimit(resource.RLIMIT_STACK, (max_memory, hard_limit))
     except (resource.error, ValueError):
-        print(error("Failed to change stack size to unlimited. Try doing it manually with `ulimit -s unlimited`."))
+        print(error(f'Failed to change stack size. Please set it to at least {max_memory // 1024}kB '
+                    f'with `ulimit -s {max_memory // 1024}`.'))
```
tonowak, line 268: Do we ever expect it to fail?

tonowak: If no, I would say that we should treat is as a bug that the user should send us a report. Currently it makes it sound like it's expected to fail sometimes.

MasloMaslane: Setting stack limit on macOS works weirdly, supposedly it can sometimes ignore soft and even hard limits. Currently using `resource.setrlimit` on my macOS always fails, so I think that on macOS we'll need to use `ulimit` to set stack limit

tonowak: Can we do it in sinol-make, instead of relying on the user to set it?

src/sinol_make/util.py
MasloMaslane: Then what should the stack size be? I thought we decided that it should be max memory limit for the task

tonowak: Even when they would use more memory than they should have? I don't think it's a good solution.

tonowak: Reminder: we want the tool to catch any programmer's mistakes. It's better for a solution to not work *now* than on sio2/during a competition.

tonowak: Right, so why do you think there is a need to set a bigger stack size than the memory limit?

MasloMaslane: Because from what I understand changing stack size is for a whole process, so if the stack size was changed for each solutions while they're running in parallel then it would be changing constantly and some solutions may have wrong stack sizes

MasloMaslane: And from my testing it seems to be the case

tonowak: Sounds good, didn't know about it. Please leave a comment explaining that somewhere.

MasloMaslane: I changed the code a bit. I also deleted the default value since we require that there is a limit for every test

tonowak (approved):

PR #105, opened 2023-08-31
Use default time and memory limit when they're not set

MasloMaslane:


tonowak (commented):
Hmm, I think it would be better to exit with an error. We *want* the users to specify their own time and memory limit. We had too much trouble with people not specifying it and thus the job of fine-tuning the limits came to the jury, which we want to avoid.

MasloMaslane:
Ok, so now the code exits with error when a limit is not set

src/sinol_make/commands/run/__init__.py
```diff
@@ -1152,14 +1152,9 @@ def run(self, args):
...
-                    package_util.get_time_limit(test, self.config, lang, self.args)
-                except KeyError:
-                    util.exit_with_error(f'Time limit was not defined for test {os.path.basename(test)} in config.yml.')
-                try:
-                    package_util.get_memory_limit(test, self.config, lang, self.args)
-                except KeyError:
-                    util.exit_with_error(f'Memory limit was not defined for test {os.path.basename(test)} in config.yml.')
+                # The functions will exit if the limits are not set
+                package_util.get_time_limit(test, self.config, lang, self.args)
+                package_util.get_memory_limit(test, self.config, lang, self.args)
```
tonowak, line 1157:
```suggestion
                _ = package_util.get_time_limit(test, self.config, lang, self.args)
                _ = package_util.get_memory_limit(test, self.config, lang, self.args)
```
Let's make it clear that the function returns a value and we are ignoring it. (I'm unsure whether the code I wrote works)

tonowak (approved):

PR #106, opened 2023-08-31
Rename `cache` directory to `.cache`

MasloMaslane:
I also added code for moving `cache` to `.cache` after updating to a newer version

### DIFF
--- a/src/sinol_make/util.py
+++ b/src/sinol_make/util.py
@@ -1,9 +1,12 @@
 import glob, importlib, os, sys, requests, yaml
 import platform
 import tempfile
+import shutil
 import hashlib
 import threading
 from typing import Union
+
+import sinol_make
 
 
 def get_commands():
@@ -263,6 +266,16 @@ def is_linux():
 def get_file_md5(path):
     with open(path, "rb") as f:
         return hashlib.md5(f.read()).hexdigest()
+
+
+def make_version_changes():
+    """
+    Function to make changes after updating sinol-make.
+    """
+    # Move `cache` directory to `.cache` directory if the version is higher than 1.5.2.
+    if compare_versions("1.5.2", sinol_make.__version__) == -1:
+        if check_if_package() and os.path.exists("cache"):
+            shutil.move(os.path.join(os.getcwd(), "cache"), os.path.join(os.getcwd(), ".cache"))
 
 
 def color_red(text): return "\033[91m{}\033[00m".format(text)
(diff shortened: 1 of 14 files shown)


src/sinol_make/util.py
```diff
@@ -265,6 +268,16 @@ def get_file_md5(path):
...
 
 
+def make_version_changes():
+    """
+    Function to make changes after updating sinol-make.
+    """
+    # Move `cache` directory to `.cache` directory if the version is higher than 1.5.2.
+    if compare_versions("1.5.2", sinol_make.__version__) == -1:
+        if check_if_package() and os.path.exists("cache"):
+            shutil.move(os.path.join(os.getcwd(), "cache"), os.path.join(os.getcwd(), ".cache"))
```
tonowak, line 278: I'm not sure I understand the `if` here. When the version is strictly higher than `1.5.2`, it moves `cache` to `.cache`?

Seeing that currently sinol-make doesn't have many users, I would say it's OK to not move and just let the user have both the `cache` and `.cache` directories, just for code simplicity. I think that people that used the old sinol-make won't use the new one on the packages that they created previously.

MasloMaslane: Sure, I deleted this part of code

tonowak (commented):
Instead of having a `".cache"` constant everywhere in the code, maybe we could do better?

tonowak:
What you think would be the cleanest solution?

MasloMaslane:
Probably a function `get_cache_path()` - returns path to cache or `get_path_in_cache(path)` - returns `./.cache/<path>`

tonowak:
Sounds reasonable. Maybe we could also do it for other constants, if there are any?

MasloMaslane:
Sure, I'll look for them

tonowak (approved):
I would say that `get_cache_path` is a bit better name than `get_path_in_cache` (and similar to e.g. executions), as the `get_path_in_cache` can suggest that we're finding a path in some cache that we're currently maintaining.

MasloMaslane:
`get_path_in_cache(paths)` returns a path inside of cache, it works like `os.path.join`

tonowak:
Right, and that's what I mean. It makes it sound like in the coffee you have some cache object, and not in the file system. The function name doesn't sound like it's appending the string "cache" to the path.

MasloMaslane:
I removed the `get_path_in_*()` functions

PR #107, opened 2023-09-01
Fix getting task id when current directory has a long name

MasloMaslane:


tonowak:
What's the context behind this PR? Why is it needed? Why the fix looks like this?

MasloMaslane:
Previously, when the directory with package had a long name (for example full task name) instead of a short name (task id), the code would work, because `get_task_id()` returned current working directory's name.
Now if current working directory's name is longer than 3 characters, the task id is retrieved from the model solution's name

tonowak:
Doesn't sio2 decide it based on the task name?

MasloMaslane:
In the archive uploaded to sio2 the task has to be in directory named with the short task id or otherwise the upload will fail

MasloMaslane:
And the task id is this directory's name

tonowak:
Let's do the same

MasloMaslane:
But the problem is that the repos with tasks usually have long names, and in CI `sinol-make` doesn't know what the task id is

MasloMaslane:
Now that I think about it, task id doesn't have to be 3 letters long, so I don't have an idea how to properly get task id

MasloMaslane:
The only solution I can think of is to have task id stored in config

tonowak:
Maybe we can change the CI, or the way we create task packages in sinol? Having the task id in the config would be *extremely confusing* as, from what I understand, it wouldn't be picked up by sio2.

MasloMaslane:
In OI's CI, as long as the task ids are 3 letters long, the code from this PR would work. The problem is when the task id would be longer.

Also if we had task id in config and `sinol-make` was used for exporting the task then it wouldn't be a huge problem. But it still could be confusing.

tonowak:
Hmm, I'm still not convinced that some heuristics for determining the task ID is the way to go. Let's try to do it in a proper way, instead of patching stuff up.

So, you say that sio2 determines the task ID based on the name of the archive?

I would be more inclined towards creating a new variable in the config (whose name suggests that it exists only for sinol-make), which is obligatory and three-letters long. It's safe to assume that task packages developed through sinol-make will be exported through sinol-make, which allows to create the archive with the correct name. Then it would also be OK to add checks for program names and test names, that is, whether they have the correct prefix. If they don't have, we could make an error or a warning and ignore them. It would probably simplify the existing code, as we would exactly know for which filenames we should look for. The errors and warnings would make it harder for the task creator to make any mistake.

What do you think?

MasloMaslane:
I agree, I think this is a good approach

src/sinol_make/helpers/package_util.py
```diff
@@ -1,12 +1,24 @@
...
+        config = yaml.load(config_file, Loader=yaml.FullLoader)
+    if "sinol_task_id" in config:
+        return config["sinol_task_id"]
+    else:
+        print(util.warning("sinol_task_id not specified in config.yml. Using task id from directory name."))
+        task_id = os.path.split(os.getcwd())[-1]
+        if len(task_id) == 3:
+            return task_id
+        else:
+            util.exit_with_error("Invalid task id. Task id should be 3 characters long.")
```
MasloMaslane, line 21: Do you think we should exit when `sinol_task_id` is not in config or try to get task id from directory's name?

tonowak: I want to avoid heuristics as much as possible. I would say that we can try to read the directory name and:
- if it's 3 letters long, assume it's the task id,
- otherwise make an error that the field is missing.

It would be an undocumented feature, but hopefully that will make sinol-make work with old packages.

What do you think?

MasloMaslane: I agree, that's a good solution

examples/config.yml
```diff
@@ -59,6 +59,12 @@ extra_compilation_files: ['abclib.cpp', 'abclib.py']
 
 ### Keys used by sinol-make:
 
+# This key is used to identify the task id. This should be the short name of the task.
+# It's important if the directory name is not the same as the task id.
+# This key is only used by `sinol-make`. If you export the package by hand remember to
+# name the directory with the task the same as the task id.
```
tonowak, line 65:
```suggestion
# This key represents the short name (consisting of 3 letters) of the task.
# The names of files in `prog/`, `doc/`, `in/` and `out/` directories have to start with this task id.
# This key is only used by `sinol-make`: running `sinol-make export` creates
# an archive with the proper name, which sio2 uses as the task id.
```

tonowak (approved):

PR #109, opened 2023-09-02
Fix compilation caching

MasloMaslane:
When a file with code A was compiled, the path to the executable was saved for md5sum of code A. Then after changing file's code to code B and compiling it, the executable's path is the same, but it's compiled with code B. The md5sum for code A is still stored. Then after changing file's code to code A again, `sinol-make` sees that there is a md5sum for code A and returns path to the executable, which is still compiled with code B.

Now, information about md5 sums for a file named `file.cpp` is stored in `.cache/md5sums/file.cpp`, which contains the md5sum of the code and path to the executable. This fixes the above problem, because after compiling the same file again, the previous md5sum is deleted.

### DIFF
--- a/src/sinol_make/helpers/compile.py
+++ b/src/sinol_make/helpers/compile.py
@@ -32,7 +32,6 @@ def check_compiled(file_path: str):
                 exe_path = info.get("executable_path", "")
                 if os.path.exists(exe_path):
                     return exe_path
-            os.unlink(info_file_path)
             return None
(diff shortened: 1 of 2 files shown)


tonowak (approved):
Interesting bug, nice catch.

Any ideas how we can secure ourselves from stumbling into a similar bug in the future?

MasloMaslane:
Only thing I can think of is to write better tests

PR #110, opened 2023-09-02
Fix running `inwer` when inwer doesn't print any output

### LINKED ISSUE #108: Code fails when inwer doesn't print any output
According to Kuba Rożek, when inwer doesn't print any output the code fails with `IndexError: list index out of range`

MasloMaslane:
Closes #108


tonowak (approved):
Any ideas how we can stop ourselves from stumbling into a similar bug in the future, or how to find similar bugs in the existing codebase?

MasloMaslane:
Probably using `sinol-make` on past tasks will be good for finding more bugs. For future I guess writing more and better tasks would help

PR #111, opened 2023-09-02
Add `doc` command

MasloMaslane:
The command works exactly the same as `makefile`'s doc command

### DIFF
--- a/.github/workflows/Arch.yaml
+++ b/.github/workflows/Arch.yaml
@@ -20,7 +20,7 @@ jobs:
       - name: Prepare system
         run: |
           sysctl kernel.perf_event_paranoid=-1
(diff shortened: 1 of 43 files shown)


tonowak (commented):
Let's wait with this PR until we'll have a meeting with Lech Duraj, or anyone else that is deeply familiar with the `tex` files we're using. I think we could replace the current statement templates with something better (which could have different compilation commands), but I'm not familiar enough with latex to think of good solutions.

tonowak (approved):
Can you test whether it works correctly on the last few tasks from OI? Does it correctly handle images and sample tests?

MasloMaslane:
I tested it on all tasks from 30 OI and they all look the same

PR #114, opened 2023-09-10
Add function for validating test names

MasloMaslane:


src/sinol_make/helpers/package_util.py
```diff
@@ -143,3 +152,30 @@ def get_memory_limit(test_path, config, lang, args=None):
 
     str_config = util.stringify_keys(config)
     return _get_limit(LimitTypes.MEMORY_LIMIT, test_path, str_config, lang)
+
+
+def validate_files(task_id):
```
tonowak, line 157: I find it to be a confusing name. Can you guess why?

src/sinol_make/helpers/package_util.py
```diff
@@ -143,3 +152,30 @@ def get_memory_limit(test_path, config, lang, args=None):
...
+            invalid = True
+            for pattern in patterns:
+                if pattern.match(os.path.basename(file)):
+                    invalid = False
+                    break
+            if invalid:
+                invalid_files.append(os.path.basename(file))
+        return invalid_files
+
+    in_test_re = re.compile(r'^(%s(([0-9]+)([a-z]?[a-z0-9]*))).in$' % (re.escape(task_id)))
```
tonowak, line 173: In general, I find regex quite unreadable. It takes a lot of time to understand this line of code and even after some time I'm not sure whether it is correct. Can we make some simple code that does what this regex tries to do, but is readable and hard to make a bug?

src/sinol_make/helpers/package_util.py
```diff
@@ -143,3 +152,30 @@ def get_memory_limit(test_path, config, lang, args=None):
 
     str_config = util.stringify_keys(config)
     return _get_limit(LimitTypes.MEMORY_LIMIT, test_path, str_config, lang)
+
+
+def validate_files(task_id):
+    """
+    Checks if all files in the package have valid names.
+    """
+    def get_invalid_files(path, *patterns):
```
tonowak, line 161: Are there multiple patterns, or only one?

src/sinol_make/helpers/package_util.py
```diff
@@ -22,20 +23,28 @@ def get_task_id() -> str:
...
+    :param task_id: Task id. If None, it is extracted from config.yml.
+                    If not found, it's assumed that the length of task id is 3.
     :return: Test group and number.
     """
-    return os.path.split(os.path.splitext(test_path)[0])[1][3:]
+    if task_id is None:
+        try:
+            task_id = get_task_id()
+        except FileNotFoundError:
+            task_id = "abc"
```
tonowak, line 39: Why `abc`?

src/sinol_make/helpers/package_util.py
```diff
@@ -22,20 +23,28 @@ def get_task_id() -> str:
...
 
 
-def extract_test_id(test_path):
+def extract_test_id(test_path, task_id=None):
     """
     Extracts test group and number from test path.
     For example for test abc1a.in it returns 1a.
     :param test_path: Path to test file.
+    :param task_id: Task id. If None, it is extracted from config.yml.
+                    If not found, it's assumed that the length of task id is 3.
```
tonowak, line 32: Maybe it would be cleaner to initialize `task_id` somewhere at the beginning of the code, and then only pass a non-None value to the functions?

src/sinol_make/helpers/package_util.py
MasloMaslane: I changed the name

src/sinol_make/helpers/package_util.py
MasloMaslane: I copied this regex from sio's code and I don't think that changing this regex to a function will be more readable

src/sinol_make/helpers/package_util.py
MasloMaslane: While I was writing this PR there was a need for multiple patterns, but now it's just one

src/sinol_make/helpers/package_util.py
tonowak: So what does this regex do, in detail?

MasloMaslane: Test name has to begin with task id, then at least one digit, then an optional letter, then any string consisting of letters and digits

PR #121, opened 2023-09-15
Multiple bugfixes for finding tests

### LINKED ISSUE #116: `run` command crashes when running only on group 0
Found by Kuba Rożek

MasloMaslane:
Bugs fixed:
- When running only on group 0 `sinol-make` would crash (Closes #116)
- When there were no tests `sinol-make` would crash
- When passing non-existing files to `--tests` flag, `sinol-make` would crash

I also added tests for these bugs


tonowak (approved):
How can we protect ourselves from similar bugs in the future? Any ideas where we still might have similar bugs?

PR #124, opened 2023-09-16
Fix `gen` command changing modification date of shell ingen

### LINKED ISSUE #122: `sinol-make` changes ingen.sh
Every time `gen` command is run, the last modification date of shell ingen changes

MasloMaslane:
Closes #122


tonowak (approved):
Maybe let's also add a test, as long as it's easy to add one?

PR #127, opened 2023-09-22
Cache run results

MasloMaslane:
I added caching test results. Also when checker is changed, all cached results are removed.


MasloMaslane:
I added new function: when a file from `extra_compilation_files` changes, all solutions in the same language as the file are removed from cache

src/sinol_make/helpers/cache.py
```diff
@@ -0,0 +1,91 @@
...
+    """
+    os.makedirs(paths.get_cache_path("md5sums"), exist_ok=True)
+    try:
+        with open(paths.get_cache_path("md5sums", os.path.basename(solution_path)), 'r') as cache_file:
+            data = yaml.load(cache_file, Loader=yaml.FullLoader)
+            if not isinstance(data, dict):
+                return CacheFile()
+            return CacheFile.from_dict(data)
+    except (FileNotFoundError, yaml.YAMLError, TypeError):
+        return CacheFile()
```
tonowak, line 24: Maybe we should make a warning about "corrupted" files?

src/sinol_make/helpers/cache.py
```diff
@@ -0,0 +1,91 @@
...
+    info.save(file_path)
+
+    if is_checker:
+        for solution in os.listdir(paths.get_cache_path('md5sums')):
+            info = get_cache_file(solution)
+            info.tests = {}
+            info.save(solution)
+
+
+def check_extra_compilation_files(extra_compilation_files, task_id):
```
tonowak, line 65: Can you try to come up with a name that more clearly shows what this function does?

src/sinol_make/structs/status_structs.py
```diff
@@ -15,3 +17,88 @@ def __str__(self):
...
+        elif status == "ML":
+            return Status.ML
+        elif status == "RE":
+            return Status.RE
+        elif status == "WA":
+            return Status.WA
+        elif status == "OK":
+            return Status.OK
+        else:
+            return Status.PENDING
```
tonowak, line 36: is there a `status` string that means `PENDING`? Maybe we should warn/error when the content of `status` doesn't make sense?

tonowak (approved):
Looks good! Some small suggestions, merge after resolving them:

PR #128, opened 2023-09-23
Add ICPC contest type

MasloMaslane:
I added `sinol_contest_type: icpc`.
I also made changes to `sinol_expected_scores` format. Before:
```yml
sinol_expected_scores:
  abc.cpp:
    points: 100
    expected: {1: "OK", 2: "OK", 3: "OK", 4: "OK"}
  abc1.cpp:
    points: 75
    expected: {1: "OK", 2: "OK", 3: "OK", 4: "WA"}
  abc2.cpp:
    points: 25
    expected: {1: "OK", 2: "WA", 3: "WA", 4: "TL"}
  abc3.cpp:
    points: 25
    expected: {1: "OK", 2: "WA", 3: "WA", 4: "ML"}
  abc4.cpp:
    points: 50
    expected: {1: "OK", 2: "OK", 3: "WA", 4: "RE"}
```
Now:
```yml
sinol_expected_scores:
  abc.cpp:
    expected:
      1: {points: 25, status: OK}
      2: {points: 25, status: OK}
      3: {points: 25, status: OK}
      4: {points: 25, status: OK}
    points: 100
  abc1.cpp:
    expected:
      1: {points: 25, status: OK}
      2: {points: 25, status: OK}
      3: {points: 25, status: OK}
      4: {points: 0, status: WA}
    points: 75
  abc2.cpp:
    expected:
      1: {points: 25, status: OK}
      2: {points: 0, status: WA}
      3: {points: 0, status: WA}
      4: {points: 0, status: TL}
    points: 25
  abc3.cpp:
    expected:
      1: {points: 25, status: OK}
      2: {points: 0, status: WA}
      3: {points: 0, status: WA}
      4: {points: 0, status: ML}
    points: 25
  abc4.cpp:
    expected:
      1: {points: 25, status: OK}
      2: {points: 25, status: OK}
      3: {points: 0, status: WA}
      4: {points: 0, status: RE}
    points: 50
```

Because of how long it became, i added a flag `--print-expected-scores` and by default `sinol-make run` only prints messages about whats changed.
This changed was required, because icpc scores didn't save properly. I suspect that for packages with `sinol_contest_type: oi` when a group got less points because of linear falling the points also hadn't saved properly.


examples/config.yml
```diff
@@ -71,9 +71,14 @@ sinol_task_id: abc
...
 # - `default` - Points for a test can only be 100 or 0 (unless checker assigns points).
 #               Points for a group are calculated based of the lowest number of points for a test in this group.
+#               If scores are not defined in `scores` key, then all groups have the same number of points,
+#                summing up to 100.
 # - `oi` - Points for a test are unchanged if the execution time is less or equal to the time limit.
 #          Otherwise, number of points decreases linearly to one depending on the execution time.
 #          Points for a group are calculated same as in `default` mode.
+# - `icpc` - A test passes, when the status is OK.
+#            A group passes, when all tests in this group pass.
+#            A solution passes, when all groups pass.
```
tonowak, line 81:
```suggestion
# - `icpc` - A test passes when the status is OK.
#            A group passes when all tests in this group pass.
#            A solution passes when all groups pass.
```

src/sinol_make/__init__.py
```diff
@@ -9,7 +9,7 @@
 from sinol_make import util, oiejq
 
 
-__version__ = "1.5.8"
+__version__ = "1.5.9"
```
tonowak, line 12: I think it's common to first merge all PRs and then create a tiny PR that just bumps the version. This approach might be useful in your case.

src/sinol_make/commands/run/__init__.py
```diff
@@ -258,9 +260,11 @@ def configure_subparser(self, subparser):
                             help=f'tool to measure time and memory usage (default: {default_timetool})')
         parser.add_argument('--oiejq-path', dest='oiejq_path', type=str,
                             help='path to oiejq executable (default: `~/.local/bin/oiejq`)')
-        add_compilation_arguments(parser)
+        parser.add_argument('-p', '--print-expected-scores', dest='print_expected_scores', action='store_true',
+                            help='print whole expected scores report')
```
tonowak, line 264: Maybe we don't need this flag at all? After all, we see how many points each solution got in the table above it.

tonowak (approved):
After resolving the comments, it's OK to merge :tada:

src/sinol_make/commands/run/__init__.py
MasloMaslane: Do you mean that we should completely abandon printing `sinol_expected_scores`? I think that they aren't needed anymore, since the table prints the same things and the messages about changes should catch all changes.

src/sinol_make/__init__.py
MasloMaslane: Sure

MasloMaslane: Sure

src/sinol_make/commands/run/__init__.py
MasloMaslane: Hmm, but it may be useful if someone wanted to compare results with config easly

Kostero:
Why do we have points here? The problems are judged binary, i.e. either it passes or not.

MasloMaslane:
The possible points are 1/0, so I thought that would be enough. But if you think that "Accepted"/"Failed" would be better I can try doing that

Kostero:
For ICPC I would imagine the following config would make the most sense:
  abc1.cpp:
    expected:
      status: OK
  abc2.cpp:
    expected:
       status: TLE

Also I am not sure if you support multiple possible statuses (for instance you might have a solution that can TLE/RE/MLE). If not, that would be a small feature request.

MasloMaslane:
It would require some changes, but I think we can support a format like that

tonowak:
@MasloMaslane can you make an issue or a PR for that?

MasloMaslane:
I made an issue

PR #129, opened 2023-09-23
Allow running `sinol-make` from package's subdirectory

MasloMaslane:


tonowak (approved):
Well, it's a bit hacky. Can we make it more clear what the function does from its name?

PR #133, opened 2023-09-25
Allow solutions with long names

MasloMaslane:


tonowak:
Does sio2 support long names?

MasloMaslane:
Not yet, but it will support them before AMPPZ

tonowak:
Let's merge it after it will support.

tonowak (approved):

PR #134, opened 2023-09-25
`inwer` improvements

MasloMaslane:


example_package/prog/abcinwer.cpp
```diff
@@ -38,13 +38,13 @@ int main() {
...
     for (auto [subtask_id, is_valid] : subtasks)
-        subtasks_s += is_valid ? to_string(subtask_id) : string("-");
+        subtasks_s += is_valid ? to_string(subtask_id) : string("_");
 
     map<string, bool> sample_tests = {
         {"0a", is_0a()},
         {"1ocen", is_1ocen()},
     };
-    string sample_test_s = "-";
+    string sample_test_s = "_";
```
MasloMaslane, line 47: These changes were suggested by Kuba Rożek. What do you think?

tonowak: I don't understand why that matters

tonowak: What's the context behind this change?

MasloMaslane: He thinks that `_` would look better than `-`

MasloMaslane: I don't know, you'd have to ask him

tonowak: I would say it's not really beneficial to make changes based solely on a single person's preferences. If that would be a change with some arguments about readability/clarity improvements, I would approve immediately. But as this change doesn't have arguments, making it is just time consuming, so I'm against. Often for each personal preference you can find someone that has the opposite personal preference, so there is a chance that someone likes the `-` variant more, thus it's not obvious at all why this change is positive.

MasloMaslane: Ok, I'll revert it

tonowak (approved):

PR #140, opened 2023-10-05
After running `sinol-make doc` remove log files

### LINKED ISSUE #139: `sinol-make doc` should clean

MasloMaslane:
Now after running `sinol-make doc` all compilation logs are removed, unless `--logs` flag is set. Closes #139

### DIFF
--- a/src/sinol_make/__init__.py
+++ b/src/sinol_make/__init__.py
@@ -9,7 +9,7 @@ import os
 from sinol_make import util, oiejq
 
 
-__version__ = "1.5.11"
+__version__ = "1.5.12"
 
 
 def configure_parsers():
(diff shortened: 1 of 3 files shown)


tonowak:
What's the reasoning behind this? Why is it needed?

Having a lot of commandline flags is sometimes harmful, as because of it the user won't necessarily read the whole `--help`.

MasloMaslane:
Compiling latex generates a lot of log files and I think that usually they aren't useful for the user and they only clutter up the `doc/` directory

tonowak:
Maybe let's them move to a cache dir, so that if needed, the user can see them?

tonowak (approved):

PR #143, opened 2023-10-09
Update install instructions in README

MasloMaslane:
Some people didn't know that `~/.local/bin` has to be in PATH

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,7 @@ through Python's package manager pip, which usually is installed alongside Pytho
 ```
 pip3 install sinol-make
 ```
+`pip` installs `sinol-make` executable in `~/.local/bin/` directory, so make sure it's in your `PATH`. 
 
 As `oiejq` works only on Linux-based operating systems,
 *we do not recommend* using operating systems such as Windows or macOS.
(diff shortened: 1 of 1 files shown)


tonowak:
I'm not sure whether it's sufficient. What if someone doesn't know what a PATH is, or how to add the path to PATH? Can we link something that explains everything?

tonowak:
Maaybee. It's more than just explaining PATH etc, as it contains some useless information about installations. To see how to add something to the PATH, the user needs to go through one more link inside of it. It would be nice to have something better, but if not, I'm fine with this.

MasloMaslane:
What about this https://gist.github.com/nex3/c395b2f8fd4b02068be37c961301caa7 ? Then readme could look like this:

`pip` installs `sinol-make` executable in `~/.local/bin/` directory, so make sure it's in your `PATH`. [Here's](https://gist.github.com/nex3/c395b2f8fd4b02068be37c961301caa7) how to add a directory to `PATH`.

tonowak:
Looks good!

Slight suggestions: 

`installs sinol-make executable in` -> `installs the sinol-make executable in the`

`so make sure it's in your` -> `so make sure this directory is in your`

tonowak (approved):

PR #144, opened 2023-10-10
Add an optional argument to fix deprecated warning in Python 3.12

MasloMaslane:
Python 3.12 added new argument to `tar.extractall` which specifies how the tar files should be handled. When this arguments is not passed `DeprecationWarning` is shown. To fix this, I added a function that passes this argument if the python version is 3.12. Here are informations about this new argument https://docs.python.org/3/library/tarfile.html#extraction-filters


tonowak:
Maybe let's always pass it? Would it work on the older version?

MasloMaslane:
On older versions code fails with this argument

tonowak:
Ugh stupid Python

tonowak (approved):

PR #155, opened 2023-10-28
Don't wait for process checking version to exit

### LINKED ISSUE #152: Main process waits for children to exit
Main process waits for the child which checks for updates, this shouldn’t happen

MasloMaslane:
The process which starts `check_version` function will now run as a daemon, so it will exit when the main process ends.

Closes #152

### DIFF
--- a/src/sinol_make/util.py
+++ b/src/sinol_make/util.py
@@ -5,7 +5,7 @@ import tempfile
 import shutil
 import hashlib
 import subprocess
-import threading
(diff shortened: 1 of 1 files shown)


tonowak (approved):

tonowak:
Are there any other places in the code where we could have a similar bug?

MasloMaslane:
No, this is the only place that we don't want to wait for child process to end

PR #156, opened 2023-10-28
Change default number of cpus

### LINKED ISSUE #153: Run on `n-1` cpu cores by default

MasloMaslane:
Now the default value for --cpus flag will be the number of cpus - 1.

Closes #153


tonowak:
What if there's only one CPU? Also, when someone have ~16 or ~32 CPUs, we could leave two or three of them. Maybe something like `available CPUs - (log2(available CPUs) - 1)`, or something like that, would work good?

MasloMaslane:
From my experience while running on all cpus when I had 10 or 24 cores I didn't experience any problems with my computer. But maybe the reason was that I wasn't running programs that were cpu intensive enough.

I think that your formula is a good idea

MasloMaslane:
I added a function for getting default cpu count and used your formula there

tonowak (approved):

Issue #154, opened 2023-10-27
Allow generating only ins/outs instead of both


A-dead-pixel:
**Is your feature request related to a problem? Please describe.**
I can't use this nice tool for generating outs from existing ins in packages without an ingen.

**Describe the solution you'd like**
Additional commands like `ingen` and `outgen` would solve this.

**Describe alternatives you've considered**
New flags for `sinol-make gen`, for example `-o, --outs` and `-i, --ins`.

MasloMaslane:
Thanks for the suggestion, added in `sinol-make v1.5.14`

PR #157, opened 2023-10-28
Allow generating only ins/outs in `gen` command

MasloMaslane:
Requested in #154. I added flags `--ins` and `--outs` for generating only inputs/outputs. Also, if for some input file an output file doesn't exist, it is generated.

### DIFF
--- a/src/sinol_make/__init__.py
+++ b/src/sinol_make/__init__.py
@@ -9,7 +9,7 @@ import os
 from sinol_make import util, oiejq
 
 
-__version__ = "1.5.13"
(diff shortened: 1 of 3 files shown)


tonowak (approved):
Looks good! One thing to consider before merging: as we have also short variants of the commands (`-i`, `-o`), maybe the longer names could be more descriptive, like `--only-inputs`? I'm open for suggestions.

MasloMaslane:
Yeah I agree, longer names would be better

tonowak (approved):

PR #180, opened 2024-02-13
Fix get_group to comply with oioioi standards

adespawn:
On sio tests in format abc1a1 belong in the test group 1. This PR should make the script consistent with oioioi when assigning test to groups .


MasloMaslane (approved):
Thanks for the fix!

PR #187, opened 2024-02-17
Fix and extend tests

adespawn:
Fix tests broken by #186 
Extend the test range on githubRunner

### DIFF
--- a/.github/workflows/GithubRunner.yaml
+++ b/.github/workflows/GithubRunner.yaml
@@ -18,8 +18,15 @@ jobs:
       - name: Install dependencies
         run: |
           pip install .[tests]
-      - name: Run pytest
+          sudo apt install texlive-latex-recommended -y
+          sudo apt install ghostscript -y
+      - name: Run github runner pytest
         env:
           PYTEST_ADDOPTS: "--color=yes"
         run: |
           python -m pytest -v --github-runner
+      - name: Run pytest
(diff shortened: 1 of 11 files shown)


adespawn:
I will also add test for feature from #186 before merging.

.github/workflows/GithubRunner.yaml
```diff
@@ -18,8 +18,15 @@ jobs:
...
+      - name: Run github runner pytest
         env:
           PYTEST_ADDOPTS: "--color=yes"
         run: |
           python -m pytest -v --github-runner
+      - name: Run pytest
+        env:
+          PYTEST_ADDOPTS: "--color=yes"
+        run: |
+          python -m pytest -v --time-tool time
```
MasloMaslane, line 32: This isn't required, as the `Ubuntu` nad `Arch` workflows already run the tests with `--time-tool time`

MasloMaslane (changes requested):

.github/workflows/GithubRunner.yaml
adespawn: Sure. I was mislead by https://github.com/sio2project/sinol-make/blob/d229512b722af2a30311df43718dbaa9c0338664/tests/conftest.py#L37 so I also fixed it

MasloMaslane (approved):

PR #188, opened 2024-02-18
Make the oiejq tests less flaky

A-dead-pixel:
This makes some test solutions run for a given amount of seconds properly under oiejq by running a loop bounded by the number of instructions executed.
On faster computers, which execute more instructions per second, the previous behaviour of looping for an amount of real time caused undesired TLE results, which made some tests fail.


MasloMaslane (approved):
Thanks for the PR, lgtm. I'll merge this after the tests pass

PR #189, opened 2024-02-18
Add support for tox and paralell pytest

A-dead-pixel:
The ability to run pytest in paralell (`-n auto`) makes running the test suite locally more bearable.
When I ran it this way with #187 and #188, there were no failures during 20 full runs, so it seems quite stable (unlike oioioi...).

As for tox, it is already used in other sio2project repositories and it spares the effort of having to setup a venv or sth.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,5 +11,7 @@ __pycache__
 src/sinol_make/data
 
 # pytest-cov
-.coverage
+.coverage*
 coverage.xml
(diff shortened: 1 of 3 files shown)


MasloMaslane (approved):
I also run the tests in parallel locally, I just never added this to requirements, especially since our runners are not powerful enough to run tests in parallel.

PR #192, opened 2024-02-19
Validate test contents

MasloMaslane:


MasloMaslane:
That's a good idea, we don't want crlf endlines.

PR #194, opened 2024-02-20
Init command

adespawn:
Create task from the given template in github repo.
Update all the places with task id to provided id

### DIFF
--- a/src/sinol_make/commands/init/__init__.py
+++ b/src/sinol_make/commands/init/__init__.py
@@ -55,9 +55,6 @@ class Command(BaseCommand):
             config.write(config_data)
 
     def run(self, args: argparse.Namespace):
-        if len(args.task_id) != 3:
-            util.exit_with_error("Invalid task id. Task id should be 3 characters long.")
-
         self.task_id = args.task_id
         destination = os.path.join(os.getcwd(), args.task_id)
         if os.path.isdir(destination) or os.path.isfile(destination):
--- a/src/sinol_make/commands/init/__init__.py
+++ b/src/sinol_make/commands/init/__init__.py
@@ -1,5 +1,6 @@
 import argparse
 import os
+import shutil
 import subprocess
 import tempfile
 
@@ -44,7 +45,7 @@ class Command(BaseCommand):
                 dest_filename = file
                 if file[:3] == 'abc':
                     dest_filename = self.task_id + file[3:]
-                os.replace(os.path.join(root, file), os.path.join(mapping[root], dest_filename))
+                shutil.move(os.path.join(root, file), os.path.join(mapping[root], dest_filename))
 
     def update_config(self):
         with open(os.path.join(os.getcwd(), 'config.yml')) as config:
--- a/src/sinol_make/commands/init/__init__.py
+++ b/src/sinol_make/commands/init/__init__.py
@@ -1,0 +1,75 @@
+import argparse
+import os
+import subprocess
+import tempfile
+
+from sinol_make import util
+from sinol_make.interfaces.BaseCommand import BaseCommand
+
+
+class Command(BaseCommand):
+    """
+    Class for "init"
+    """
+
+    def get_name(self):
+        return "init"
+
+    def configure_subparser(self, subparser: argparse.ArgumentParser):
+        parser = subparser.add_parser(
+            self.get_name(),
+            help='Create package from the template',
+            description='Create package from predefined template. You have specify new task shortcut (like: '
+                        'sinol-make init abc )'
+        )
+        parser.add_argument('task_id', type=str, help='Id of the task to create')
+
+    @staticmethod
+    def download_template():
+        repo = 'https://github.com/sio2project/sinol-make.git'
+        package_dir = 'sinol-make/example_package'
+        tmp_dir = tempfile.mkdtemp()
+        subprocess.run(['git', 'clone', repo], cwd=tmp_dir, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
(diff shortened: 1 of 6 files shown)


src/sinol_make/commands/init/__init__.py
```diff
@@ -0,0 +1,75 @@
...
+        )
+        parser.add_argument('task_id', type=str, help='Id of the task to create')
+
+    @staticmethod
+    def download_template():
+        repo = 'https://github.com/sio2project/sinol-make.git'
+        package_dir = 'sinol-make/example_package'
+        tmp_dir = tempfile.mkdtemp()
+        subprocess.run(['git', 'clone', repo], cwd=tmp_dir, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
+        return os.path.join(tmp_dir, package_dir)
```
A-dead-pixel, line 33: I think there should be better error handling here, i.e. git's stderr should be directed to our stderr with subprocess.PIPE and we should check whether this command succeeds, otherwise print an error and exit cleanly.

I would also add `--depth 1` and maybe `-q` to the git command.

adespawn: The output of subprocesses should be automatically redirected to main process stdout/err.

src/sinol_make/commands/init/__init__.py
```diff
@@ -0,0 +1,78 @@
...
+        with open(os.path.join(os.getcwd(), 'config.yml')) as config:
+            config_data = config.read()
+        config_data = config_data.replace('sinol_task_id: abc', f'sinol_task_id: {self.task_id}')
+
+        with open(os.path.join(os.getcwd(), 'config.yml'), 'w') as config:
+            config.write(config_data)
+
+    def run(self, args: argparse.Namespace):
+        if len(args.task_id) != 3:
+            util.exit_with_error("Invalid task id. Task id should be 3 characters long.")
```
MasloMaslane, line 59: I don't think we should limit the task id to exactly 3 characters.

src/sinol_make/commands/init/__init__.py
```diff
@@ -0,0 +1,78 @@
...
+                    dest_filename = self.task_id + file[3:]
+                os.replace(os.path.join(root, file), os.path.join(mapping[root], dest_filename))
+
+    def update_config(self):
+        with open(os.path.join(os.getcwd(), 'config.yml')) as config:
+            config_data = config.read()
+        config_data = config_data.replace('sinol_task_id: abc', f'sinol_task_id: {self.task_id}')
+
+        with open(os.path.join(os.getcwd(), 'config.yml'), 'w') as config:
+            config.write(config_data)
```
MasloMaslane, line 55: I'm not sure about it, but maybe we should also change the title? This would require adding a flag for the title.

MasloMaslane (changes requested):

src/sinol_make/commands/init/__init__.py
adespawn: Task id is already required to be exactly 3 characters long:
https://github.com/sio2project/sinol-make/blob/0ed77a26199ddfebf70bb33105d221c46cbe9353/src/sinol_make/helpers/package_util.py#L20-L23

src/sinol_make/commands/init/__init__.py
adespawn: The idea was to simplify process of creating new package, by using the template and updating taskid in all files and filenames. In my opinion providing title in the command would be a bit unnecessary, especially because other fields in config also have to be updated by the person creating the task (like limits, subtasks, ...).

MasloMaslane: Thats a good point

src/sinol_make/commands/init/__init__.py
MasloMaslane: That's only true if `sinol_task_id` is not set in config. `sinol_task_id` can be any length

A-dead-pixel:
I just ran this locally and got the following traceback:
```
Traceback (most recent call last):
  File "/home/pixl/.local/share/pipx/venvs/sinol-make/lib/python3.11/site-packages/sinol_make/__init__.py", line 72, in main
    main_exn()
  File "/home/pixl/.local/share/pipx/venvs/sinol-make/lib/python3.11/site-packages/sinol_make/__init__.py", line 64, in main_exn
    command.run(args)
  File "/home/pixl/.local/share/pipx/venvs/sinol-make/lib/python3.11/site-packages/sinol_make/commands/init/__init__.py", line 72, in run
    self.move_folder()
  File "/home/pixl/.local/share/pipx/venvs/sinol-make/lib/python3.11/site-packages/sinol_make/commands/init/__init__.py", line 47, in move_folder
    os.replace(os.path.join(root, file), os.path.join(mapping[root], dest_filename))
OSError: [Errno 18] Invalid cross-device link: '/tmp/tmpt5o6kj_l/sinol-make/example_package/.gitignore' -> '/home/pixl/sus/.gitignore'
```
On my system /tmp is on a different filesystem than /.

src/sinol_make/commands/init/__init__.py
adespawn: Good point. Sometimes understanding read code/text is quite hard 😅

adespawn:
> On my system /tmp is on a different filesystem than /, which is the cause of this error.

Didn't think about that. This (https://github.com/sio2project/sinol-make/pull/194/commits/e3dff0c93f06eecff247333752576427211375bc) should fix this problem.

MasloMaslane (approved):

PR #195, opened 2024-02-20
Add contest specific arguments

adespawn:
Make ocen creating based on a flag


adespawn:
The idea is to increase code readability and allow other contests to also have ocen export.

MasloMaslane (approved):

Issue #185, opened 2024-02-15
Latex compilation


adespawn:
Is the current way of compiling latex intended? 
From what I can see the oioioi uses ``pdflatex`` [  https://github.com/sio2project/oioioi/blob/6c9328e0ec79e3cc545c182e28f55dc1ab9ba931/oioioi/base/utils/pdf.py#L30 ]
While the script compiles the pdf with ``latex`` to ``*.dvi`` and then with ``dvipdf`` to pdf

MasloMaslane:
The current way was copied from old OI's jury makefiles. The function you linked isn't actually used when generating problem statements. oioioi only compiles latex when `USE_SINOLPACK_MAKEFILES` is set to `True` in settings.py, then it uses the makefiles which I think compile latex files in the same way as sinol-make.

adespawn:
I see... If it is done as you described, it feels a bit obscure way of doing it, but at least there is no parity issues

MasloMaslane:
If you know that `pdflatex` works better than the current way, then you can change it. Most oioioi instances shouldn't have `USE_SINOLPACK_MAKEFILES = True` as its quite insecure.

adespawn:
Ok, I will look into it. 

I brought this up, because when [creating template style](https://github.com/Stowarzyszenie-Talent/st-make/commit/418e12b8a23740c554d3e6914e072e0bd5136f25) I had problems with logo when using current method while it worked with  ``pdflatex``

adespawn:
From what i found, the ``pdflatex`` is just a simpler and a bit improved way of compiling ``.tex`` files than the current method used in the script. 

The only maior difference between the two methods  is the support for images:
``Pdflatex supports .png, .jpg and .pdf images`` 
while ``original TeX and DVI format have support for Encapsulated PostScript files (.eps)``
[according to this article](https://www.overleaf.com/learn/latex/Articles/The_TeX_family_tree%3A_LaTeX%2C_pdfTeX%2C_XeTeX%2C_LuaTeX_and_ConTeXt)

On top of that it's easier to install ``pdflatex`` than (``latex`` and ``dvipdf``) as the second one requires additional package to be installed for ``dvipdf`` 

Unless this script will be used with statements that use ``.eps`` images I will implement this change. 
Otherwise this can be implemented as an optional feature (for example the statement method can depend on package config.yml or the flag the sinol-make is run with)

MasloMaslane:
We are using .ps images. Do they work with pdflatex?

[...]

PR #197, opened 2024-02-22
Update statement compilation

adespawn:
Add document compilation with pdflatex to allow standard image types (jpg, png) to be included in the statement. (see #185) 
Keep old compilation as an option.


src/sinol_make/commands/doc/__init__.py
```diff
@@ -56,11 +70,28 @@ def configure_subparser(self, subparser: argparse.ArgumentParser):
             help='Compile latex files to pdf',
             description='Compiles latex files to pdf. By default compiles all files in the `doc` directory.\n'
                         'You can also specify files to compile.')
+        parser.add_argument('--latex-compiler', dest='latex_compiler', choices=['auto', 'pdflatex', 'latex_dvi'],
+                            help='Compiler used to compile documents (default: auto)', default='auto')
```
MasloMaslane, line 74: Maybe you could add an explanation which compiler should be used and what `auto` does?

tests/packages/ps_doc/doc/doczad.tex
MasloMaslane, line 1: Will the tests pass? Shouldn't the files be called `doc2zad.tex` since that's the task id in config?

MasloMaslane (changes requested):

tests/packages/ps_doc/doc/doczad.tex
adespawn: The command doc does not care for task id, so the tests work perfectly (see the job status on the source branch). When not provided with file, compiles all ``.tex`` files.

src/sinol_make/commands/doc/__init__.py
adespawn: Directly in help field, or in other way?

MasloMaslane: In the help field

MasloMaslane (approved):

PR #198, opened 2024-02-26
Better indicate availability of new version.

adespawn:
Add sleep to increase visibility of information about new versions.
Comply with PEP 8: E501 and PEP 8: E722

### DIFF
--- a/src/sinol_make/__init__.py
+++ b/src/sinol_make/__init__.py
@@ -32,6 +32,21 @@ def configure_parsers():
     return parser
 
 
+def check_oiejq():
+    if util.is_linux() and not oiejq.check_oiejq():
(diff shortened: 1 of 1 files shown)


A-dead-pixel:
I would personally be annoyed by a 2-second sleep.
Imo it would be better to move the check just below `command.run(args)` or even outside of `main_exn`, so that the info would be printed even when something is wrong and the command exits.

Then the warning should always be visible at the bottom of the output and wouldn't inconvenience the user as much.

MasloMaslane:
Maybe after this line? https://github.com/sio2project/sinol-make/blob/025c88cdcabfc24840382aea0f325904e189ccdf/src/sinol_make/__init__.py#L76

Or even completely at the end of `main()`

MasloMaslane (approved):

PR #200, opened 2024-02-28
Additional compilation flags

adespawn:


MasloMaslane (changes requested):
I don't really like the idea of completely deleting the `--weak-compilation-flag`. Maybe you could create an alias or something like that?

adespawn:
Wouldn't it be quite confusing to have two separate flags for the same functionality?

MasloMaslane:
Maybe. The reason I won't to preserve `--weak-compilation-flags` is because some people may already be used to it's shortcut (`-W`) and most old packages don't compile with the current strict flags. I think that there should at least be a way to use this flag with fewer characters than `--compile-mode weak` (maybe `-c w`?). What are your thoughts @tonowak

adespawn:
Giving shorter aliases for this flag seems to me as a better option than keeping the ``--weak-compilation-flags``. 
Before I change this, how do you feel about following aliases:
``--compile-mode``: ``-c``
``default``: ``d``
``oioioi``: ``o``
``weak``: ``w``
Eventually without ``d`` alias as it's a default one. 

When it comes to the old packages, this PR aims to improve support for them, as oioioi uses some of the flags (those included in the option oioioi) and packages failing to compile with those flags *shouldn't* be possible to upload to sio2 site.

MasloMaslane:
> Giving shorter aliases for this flag seems to me as a better option than keeping the `--weak-compilation-flags`. Before I change this, how do you feel about following aliases: `--compile-mode`: `-c` `default`: `d` `oioioi`: `o` `weak`: `w` Eventually without `d` alias as it's a default one.
> 
> When it comes to the old packages, this PR aims to improve support for them, as oioioi uses some of the flags (those included in the option oioioi) and packages failing to compile with those flags _shouldn't_ be possible to upload to sio2 site.

I think that's a good solution

MasloMaslane (approved):

MasloMaslane (approved):

PR #213, opened 2024-03-09
Fix macOS workflow

MasloMaslane:


adespawn:
It appears the problem with failing macOS tests is not the problem with the given job, but with the worker...
[In our fork](https://github.com/Stowarzyszenie-Talent/st-make/actions/workflows/macOS.yaml) we have the same problem, and there is no significant difference between commit that passes the job and the one failing it

PR #221, opened 2024-03-21
Remove fsanitize for ingen and inwer

MasloMaslane:
Some linux distros have `vm.mmap_rnd_bits` set by default to 32. This sometimes breaks address and UB sanitizers. A workaround for this is to set `sudo sysctl vm.mmap_rnd_bits = 28`, but we decided that we'd rather delete sanitizers for now. The code is not completely deleted, because in future a `verify` command will run ingen and inwer with proper warnings and those sanitizers enabled

### DIFF
--- a/src/sinol_make/commands/ingen/ingen_util.py
+++ b/src/sinol_make/commands/ingen/ingen_util.py
@@ -47,7 +47,7 @@ def get_ingen(task_id, ingen_path=None):
     return correct_ingen
 
 
-def compile_ingen(ingen_path: str, args: argparse.Namespace, compilation_flags='default'):
+def compile_ingen(ingen_path: str, args: argparse.Namespace, compilation_flags='default', use_fsanitize=False):
     """
     Compiles ingen and returns path to compiled executable.
     If ingen_path is shell script, then it will be returned.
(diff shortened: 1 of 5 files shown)


A-dead-pixel:
I found that setting the stack limit with `ulimit -s` to some large value (I tried values from 1e3 to 3e7) instead of `unlimited` fixes the issue, assuming that we're talking about the same ASan error.
The traceback for me starts with the following:
```
==9508==Shadow memory range interleaves with an existing memory mapping. ASan cannot proceed correctly. ABORTING.
==9508==ASan shadow was supposed to be located in the [0x00007fff7000-0x10007fff7fff] range.
==9508==This might be related to ELF_ET_DYN_BASE change in Linux 4.12.
==9508==See https://github.com/google/sanitizers/issues/856 for possible workarounds.
```

MasloMaslane:
Yes, we are getting the same issue. We'll test if your fix works

MasloMaslane:
Unfortunately, this didn't fix the issue for everyone, only for some people. It seems that an only consistent way to fix this is to set `vm.mmap_rnd_bits `

MasloMaslane:
Sanitizers were changed in #225

PR #222, opened 2024-03-22
Autoupdate oiejq

MasloMaslane:


src/sinol_make/oiejq/__init__.py
```diff
@@ -14,13 +14,21 @@ def _check_if_oiejq_executable(path):
         return False
 
     try:
-        p = subprocess.Popen([path], shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
-        p.wait()
-        return p.returncode == 0
+        oiejq = subprocess.Popen([path], shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        oiejq.wait()
+        return oiejq.returncode == 0
     except FileNotFoundError:
```
A-dead-pixel, line 20: Is this really needed?
When testing outside of sinol-make, I couldn't trigger an exception.

MasloMaslane: That's true, in theory here an exception shouldn't happen. I'll delete it

A-dead-pixel (changes requested):
Firstly, maybe we could future-proof this somewhat and also account for updates to the oiejq script itself?
Comparing `hashlib.md5(open(path,'rb').read()).hexdigest()` agains a saved md5 checksum should suffice.

Secondly, when oiejq is updated the user sees the following:
```
`oiejq` in `~/.local/bin/` not found, installing now...
`oiejq` was successfully installed.
```
I think that this is a bit misleading and may confuse the user.
If we don't want to properly differentiate between installing and updating oiejq then I'd suggest outputting sth like 
```
Up to date `oiejq` in `~/.local/bin/` not found, installing new version...
```

MasloMaslane:
Does are good ideas, thanks

A-dead-pixel (changes requested):
Can you add a few lines to test "updating" the oiejq script after e.g. appending '\n' to it?

A-dead-pixel (approved):

PR #225, opened 2024-04-04
`verify` command

MasloMaslane:


adespawn:
Would it be a good idea to clear all cache before executing the verify? There is still some bugs with the cache not updating properly and sometimes it may become corrupted. IMO it would be more worth it to have extra confidence at the cost of the extra runtime.

MasloMaslane:
Yes, that's a good idea

src/sinol_make/commands/export/__init__.py
```diff
@@ -109,12 +108,27 @@ def create_ocen(self, target_dir: str):
...
                     shutil.copy(test, os.path.join(ocen_dir, ext, os.path.basename(test)))
-
-            shutil.make_archive(os.path.join(attachments_dir, f'{self.task_id}ocen'), 'zip', tmpdir)
+                    num_tests += 1
+
+            dlazaw_dir = os.path.join(os.getcwd(), 'dlazaw')
+            if num_tests == 0:
+                print(util.warning('No ocen tests found.'))
+            elif os.path.exists(dlazaw_dir):
+                print(util.warning('Skipping ocen arhive creation because dlazaw directory exists.'))
```
adespawn, line 122: archive

src/sinol_make/commands/verify/__init__.py
```diff
@@ -0,0 +1,173 @@
...
+                        'Ingen and inwer are compiled with sanitizers (-fsanitize=address,undefined), '
+                        'which may fail on some systems. To fix this, run `sudo sysctl vm.mmap_rnd_bits = 28` '
+                        'or disable sanitizers with --no-fsanitize.'
+        )
+
+        parser.add_argument('-e', '--expected-contest-type', type=str,
+                            help='expected contest type. Fails if the actual contest type is different.')
+        parser.add_argument('-f', '--no-fsanitize', action='store_true', default=False,
+                             help='do not use sanitizers for ingen and inwer programs')
+        parser.add_argument('-c', '--cpus', type=int,
```
adespawn, line 41: Shouldn't `add_cpus_argument` be used here?

src/sinol_make/helpers/parsers.py
```diff
@@ -20,12 +20,20 @@ def add_compilation_arguments(parser: argparse.ArgumentParser):
...
     parser.add_argument('--compile-mode', '-C', dest='compile_mode', choices=['default', 'oioioi', 'weak', 'd', 'o', 'w'],
                         help='Warning flag groups used to compile C/C++ files. Available options:\n'
                              ' default / d - uses default flags: \n'
                              '    (-Wshadow -Wconversion -Wno-unused-result -Wfloat-equal) + oioioi flags\n'
                              ' oioioi / o - uses the same flags as oioioi:\n'
                              '    (-Wall -Wno-unused-result -Werror)'
                              ' weak / w - disable all warning flags during C and C++ compilation', default='default')
+
+
+def add_cpus_argument(parser: argparse.ArgumentParser, help: str):
```
adespawn, line 35: Some other flags, like `fsanitize`, are also used in a few places. Maybe also add helper functions for those arguments?

PR #233, opened 2024-04-05
Support for interactive tasks via IO

MasloMaslane:
TODO:
- [x] support in oiejq
- [x] cache handling
- [x] don't run outgen
- [ ] refactor


adespawn:
Maybe some support for two or more execution interactive tasks?

adespawn:
Currently I'm aware of 2 implementation of what I call multi-phase (or Encoder-decoder) interactive tasks : https://github.com/sio2project/oioioi/pull/191 and https://github.com/Stowarzyszenie-Talent/sioworkers/commit/b9948b8421d8fe3cc11c11397d6518d686593afc . I don't know anything about the implementation for OI, but sometimes I create tasks using the second implementation. It would be nice if sinol-make have some support for those type of tasks. Either by formalizing the format of those tasks and implementing it into sinol-make, or for ex. allowing to create custom bash script that simulate the execution on the sio and are run by sinol-make instead of binary.
Now I realize this is not related to IO interactive tasks, but just to the interactive tasks... It's hard for me to tell which of the options would be better for OI and related competitions (or even if it will be needed at all) but if you would prefer the second option, I could look into it next week.

adespawn:
And I said something without looking into the code. And there is the the task_type interface you just implemented, that could support what I just said...

src/sinol_make/commands/run/__init__.py
```diff
@@ -631,18 +635,16 @@ def run_solution(self, data_for_execution: ExecutionData):
         Run an execution and return the result as ExecutionResult object.
         """
 
-        (name, executable, test, time_limit, memory_limit, timetool_path, execution_dir) = data_for_execution
+        (name, executable, test, output_test, time_limit, memory_limit, timetool_path, execution_dir) = data_for_execution
         file_no_ext = paths.get_executions_path(name, package_util.extract_test_id(test, self.ID))
         output_file = file_no_ext + ".out"
         result_file = file_no_ext + ".res"
-        hard_time_limit_in_s = math.ceil(2 * time_limit / 1000.0)
+        hard_time_limit = math.ceil(2 * time_limit / 1000.0)
```
adespawn, line 642: Is this 2 times time limit in any way tested to be enough? I have an example, where sio2jail runs 1.5 times longer than time measured by the jail. If not, given "worse" program or slower machine, this hard time limit may be exceeded, without exceeding time_limit.

MasloMaslane: It seems to work well in real sioworkers. But I guess on some systems it may fail

A-dead-pixel: This may be another discrepancy between upstream sioworkers and those on sio2.mimuw, but the first repo contains a multiplier of ***16*** with an additional second, see [here](https://github.com/sio2project/sioworkers/blob/9137c6a3726dd6de422dc79ed16319d782ec78f4/sio/workers/executors.py#L598-L599) and [here](https://github.com/sio2project/sioworkers/blob/9137c6a3726dd6de422dc79ed16319d782ec78f4/sio/workers/executors.py#L619-L627).

MasloMaslane: Okay, I didn't know about this. On sio2.mimuw we also have 16x time limit for sio2jail

PR #242, opened 2024-07-19
Interactive via IO support

MasloMaslane:
Missing tests, but otherwise everything should work

### DIFF
--- a/src/sinol_make/executors/time.py
+++ b/src/sinol_make/executors/time.py
@@ -16,11 +16,11 @@ class TimeExecutor(BaseExecutor):
         if sys.platform == 'darwin':
             time_name = 'gtime'
         elif sys.platform == 'linux':
-            time_name = 'time'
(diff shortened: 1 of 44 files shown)


src/sinol_make/executors/time.py
```diff
@@ -16,11 +16,11 @@ def _wrap_command(self, command: List[str], result_file_path: str) -> List[str]:
         if sys.platform == 'darwin':
             time_name = 'gtime'
         elif sys.platform == 'linux':
-            time_name = 'time'
+            time_name = '/usr/bin/time'
```
A-dead-pixel, line 19: Can we refrain from hardcoding paths to system binaries?
If the only reason for this was that we want to use the GNU time instead of a shell one, then '\time' should suffice.

MasloMaslane: fixed

PR #265, opened 2024-12-08
Use sio2jail on WSL

j4b6ski:
WSL2 (released 2019) is based on a Linux VM. In general its behavior is consistent with any other Linux system running in a VM, so there should be no reason to discriminate WSL under `util.is_linux()`.

In the future, we might want to check for specific features such as perf counters being supported, but that is an orthogonal concern.

I have tested with several people using WSL across relatively modern laptops (Intel 8th gen and up) and in all cases the Intel PMU driver passed through enough perf counters for sio2jail to work.

If we wanted to test support for sio2jail more thoroughly, I suspect the best method might be to actually try running sio2jail or the perf tool.

### DIFF
--- a/src/sinol_make/__init__.py
+++ b/src/sinol_make/__init__.py
@@ -38,7 +38,7 @@ def configure_parsers():
 
 
 def check_sio2jail():
-    if util.is_linux() and not sio2jail.check_sio2jail():
+    if sio2jail.sio2jail_supported() and not sio2jail.check_sio2jail():
         print(util.warning('Up to date `sio2jail` in `~/.local/bin/` not found, installing new version...'))
         try:
             if sio2jail.install_sio2jail():
--- a/src/sinol_make/commands/run/__init__.py
+++ b/src/sinol_make/commands/run/__init__.py
@@ -280,7 +280,7 @@ class Command(BaseCommand):
                 After running the solutions, it compares the solutions\' scores with the ones saved in config.yml.'
         )
 
-        default_timetool = 'sio2jail' if util.is_linux() else 'time'
+        default_timetool = 'sio2jail' if sio2jail.sio2jail_supported() else 'time'
 
         parser.add_argument('-s', '--solutions', type=str, nargs='+',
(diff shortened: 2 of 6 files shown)


MasloMaslane:
> In the future, we might want to check for specific features such as perf counters being supported, but that is an orthogonal concern.

There is a test if perf counters are enabled, isn't there? I don't remember exactly how it works, so I may be wrong:

https://github.com/sio2project/sinol-make/blob/1e5a62c85da638904e0b107d2f9988bd0680b130/src/sinol_make/sio2jail/__init__.py#L75

j4b6ski:
> > In the future, we might want to check for specific features such as perf counters being supported, but that is an orthogonal concern.
> 
> There is a test if perf counters are enabled, isn't there? I don't remember exactly how it works, so I may be wrong:
> 
> https://github.com/sio2project/sinol-make/blob/1e5a62c85da638904e0b107d2f9988bd0680b130/src/sinol_make/sio2jail/__init__.py#L75

Ah, that's right, that's perfect. 
It is probably a bit too expensive to put inside `sio2jail_supported()`.
But I can see it's already called before `sinol-make run` which seems like a good compromise.

j4b6ski:
My only issue with `check_perf_counters_enabled()` is the comment/help text is not entirely accurate.

From my testing, setting `kernel.perf_event_paranoid=2`, which allows accessing perf counters on the user's processes, is enough. Note that this is the maximum documented setting, but at least ubuntu/debian provide an even more restrictive default of `3`, which no longer allows perf at all.

This sio2jail check could also fail with some other issue, such as lack lacking intel PMU driver entirely, or it not supporting enough perf counters. In which case changing `perf_event_paranoid` won't help.

j4b6ski:
New diagnostic, depending on perf_event_paranoid setting:
{CCACAF99-A796-410F-9361-A0F005ACB503}
{953D8001-9CA9-483B-85DB-67AAB3FE0882}

MasloMaslane (approved):

PR #269, opened 2024-12-08
Add flag to choose template

j4b6ski:
This change allows providing a custom template arg flag -t/--template:
```
sinol-make init foo -t [repo link or disk path] [optional subdir]
```
It defaults to the upstream repository and example_package subdir.

Changed the example_package id pattern from `abc` to `__ID__`. All filenames and file contents are now substituted.

Changed the positional output directory argument to `-o/--output` flag to help avoid accidents.

Added result directory cleanup on failure.


src/sinol_make/commands/init/__init__.py
```diff
@@ -23,21 +27,35 @@ def configure_subparser(self, subparser: argparse.ArgumentParser):
...
+        subdir = template_paths[1] if len(template_paths) > 1 else ''
+
         self.used_tmpdir = tempfile.TemporaryDirectory()
         tmp_dir = self.used_tmpdir.name
-        ret = subprocess.run(['git', 'clone', '-q', '--depth', '1', repo], cwd=tmp_dir)
-        if ret.returncode != 0:
-            util.exit_with_error("Could not access repository. Please try again.")
-        path = os.path.join(tmp_dir, package_dir)
+
+        is_url = template.startswith("https://") or template.startswith("git@")
```
MasloMaslane, line 47: What about `ssh://`?

src/sinol_make/commands/init/__init__.py
```diff
@@ -80,13 +106,17 @@ def run(self, args: argparse.Namespace):
...
+            self.move_folder()
+            self.update_task_id()
 
-        self.used_tmpdir.cleanup()
+            self.used_tmpdir.cleanup()
 
-        print(util.info(f'Successfully created task "{self.task_id}"'))
+            print(util.info(f'Successfully created task "{self.task_id}"'))
+        except:
+            shutil.rmtree(destination)
```
MasloMaslane, line 121: Shouldn’t `self.used_tmpdir` also be cleaned up?

MasloMaslane (changes requested):

src/sinol_make/commands/init/__init__.py
j4b6ski: I didn't realize other schemes existed 🤔 
Added some more to the list

src/sinol_make/commands/init/__init__.py
j4b6ski: It should be, thanks
Moved that cleanup into finally block

src/sinol_make/commands/init/__init__.py
```diff
@@ -23,21 +27,32 @@ def configure_subparser(self, subparser: argparse.ArgumentParser):
...
+        print(('Cloning' if is_url else 'Copying') + ' template ' +
+            (f'{subdir} from {template}' if subdir else f'{template}'))
+        if is_url:
+            ret = subprocess.run(['git', 'clone', '-v' if verbose else '-q', '--depth', '1', template, tmpdir])
+            if ret.returncode != 0:
+                util.exit_with_error("Could not access repository. Please try again.")
+            path = os.path.join(tmpdir, subdir)
+        else:
+            path = os.path.join(tmpdir, 'template')
+            shutil.copytree(os.path.join(os.getcwd(), template, subdir), path)
```
MasloMaslane, line 54: why is there  `os.getcwd()`? Can't `template` be an absolute path?

MasloMaslane, line 54: why is there `os.getcwd()`? Can't `template` be an absolute path?

MasloMaslane (changes requested):

src/sinol_make/commands/init/__init__.py
j4b6ski: seems like os.path.join is surprisingly clever and if template here is an absolute path then os.getcwd() is ignored entirely
image
but os.getcwd() was not required here, so I removed it and added some tests to make sure both absolute and relative paths work

MasloMaslane (approved):

PR #280, opened 2025-01-19
Run sio2jail self-check with Sio2jailExecutor

j4b6ski:
We currently don't pass enough cmdline arguments to sio2jail self-check,
which makes it fail erroneously on some system configurations.
This change prevents that from happening, by using Sio2jailExecutor to
run the self check.

Notably this change required adding the original cmdline to
ExecutionResult for logging errors later - this feels like a leaky
abstraction, but that class actually already carries the process exit
signal and stderr.

Also improved sio2jail related error messages for better user experience.

### DIFF
--- a/src/sinol_make/__init__.py
+++ b/src/sinol_make/__init__.py
@@ -12,7 +12,7 @@ from sinol_make.task_type.normal import NormalTaskType # noqa
 from sinol_make.task_type.interactive import InteractiveTaskType # noqa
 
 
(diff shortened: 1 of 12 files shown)


j4b6ski:
improved error messages
Screenshot 2025-01-19 154842
Screenshot 2025-01-19 154902
Screenshot 2025-01-19 155829

MasloMaslane (approved):

PR #302, opened 2025-07-20
Enhance static file handling and add globbing support in tests

### LINKED ISSUE #281: Support globbing in sinol_static_tests
**Is your feature request related to a problem? Please describe.**
Forgetting to update `sinol_static_tests` after creating a testcase manually can result in annoying data loss.

**Describe the solution you'd like**
I believe this can be significantly mitigated by encoding the naming convention itself into the configuration file. For example,

```yaml
sinol_static_tests:
- abc0*.in
- abc?ocen.in
```

**Describe alternatives you've considered**
An orthogonal issue would be tracking the actual output of `ingen` somewhere within `.cache`, and requiring confirmation when cleaning up a file that isn't specified in `sinol_static_tests`, but also we have no record of it being created in a previous run of `ingen`.

MasloMaslane:
Closes #281

### DIFF
--- a/example_package/config.yml
+++ b/example_package/config.yml
@@ -81,7 +81,8 @@ sinol_contest_type: oi
 sinol_latex_compiler: auto
 
 # You can specify which tests are static (handwritten). This allows sinol-make to differentiate between
-# old and handwritten tests. If this key is not present old tests won't be removed.
+# old and handwritten tests. If this key is not present old tests won't be removed. The values can be
+# either a list of test names (e.g. `["__ID__0.in", "__ID__1.in"]`) or a glob pattern (e.g. `"__ID__*.in"`).
 # This key is optional and should be a list of tests.
 sinol_static_tests: ["__ID__0.in", "__ID__0a.in"]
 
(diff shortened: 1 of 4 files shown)


example_package/config.yml
```diff
@@ -81,7 +81,8 @@ sinol_contest_type: oi
 sinol_latex_compiler: auto
 
 # You can specify which tests are static (handwritten). This allows sinol-make to differentiate between
-# old and handwritten tests. If this key is not present old tests won't be removed.
+# old and handwritten tests. If this key is not present old tests won't be removed. The values can be
+# either a list of test names (e.g. `["__ID__0.in", "__ID__1.in"]`) or a glob pattern (e.g. `"__ID__*.in"`).
```
meithecatte, line 85: This suggests that only one glob pattern can be used. Reading the code suggests that multiple glob patterns are handled just fine, but it'd be nice to also make that clear in the documentation.

MasloMaslane: https://github.com/sio2project/sinol-make/commit/b61cab7e7c748b551a6d56141c054ccbcfba14ba

